PR #10, opened 2022-10-10
Feature/dread

SondreHus:


Fueredoriku (approved):
I don't know how I feel about Bob, but I approve of the library choices 👍

Assets/Animation/bob.txt
```diff
@@ -0,0 +1,5 @@
+Hi, I'm Bob, i cease to exist once you delete me. My existance is brief, but it was nice to meet you.
+
+Till' next time
```
toberge, line 3: see you later bob
also why Animation and not Animation**s**

PR #12, opened 2022-10-12
Feature/input manager

toberge:
- Add input manager that uses Unity's Player Input Manager components to
  provide an easy interface of delegates to other scripts.
- Delete our friend `bob.txt` twice :(


Fueredoriku (approved):
Updated Eventlisteners to have the callbackContext as parameter when invoked.

Everything looks good and works on my machine, and I very much approve of the overall structure.  The only downside of the implementation being exceptions in the console for invoking delegates with no hooks attached, but this won't be a problem long-term anyways.

On an unrelated note, I think we should probbly add ProjectSettings/EditorBuildSettings.asset and ProjectSettings/EditorBuildSettings.asset to gitignore 

Anyways, I approve, good work! 😎

toberge:
I'm not sure why `blahblah.completed += OnWhatever;` doesn't work by itself and you need to wrap it in `(ctx) => OnWhatever(ctx)`, but I suppose it's just the type system throwing a fit?

These two commits should be squashed IMO:
image

On another unrelated note, please limit commit message headers to 50 characters, this is hard to read:  
image

I suggest writing it more like this (body should be ~72 chars wide or sth, Vim wraps it for me):

```
Update input manager to invoke events with context

- Updated eventlisters to invoke inputEvents with callback context as a parameter.
- Moved override eventlistener function call inside the relevant add/remove eventlistener functions.
 This implementations has the downside of giving error messages when no functions are added to
  an inputEvent and the inputEvent is invoked, but this isn't relevant for our implementation as all
  these delegates will have functions later.
```

This is why one might want stricter Git guidelines :)))))))

toberge:
& note that I cannot merge this myself btw

Fueredoriku:
> but I suppose it's just the type system throwing a fit?

Usually it works according to [the documentation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/delegates#196-delegate-invocation), but I think it specifically doesn't work in this situation because we're hooking a delegate to a delegate instead of hooking a function to a delegate.

PR #22, opened 2022-11-02
Match manager

Fueredoriku:
A component for managing a match
- Every match has multiple rounds
  - The player with 3 rounds won, wins the entire match
  - Round win condition is last person standing
- Added currency "chips"
  - Chips are rewarded at end of every round
  - You get 1 chip as base each round, 2 chips for winning, and 1 chip for each player you killed during a round
- Players are sent back to menu after match is complete.

//TODO:
- Add biddingRound functionality after bidding PR is merged
- Remove debug.logs + document code

PR co-written by @toberge


Assets/Scripts/Gamestate/MatchController.cs
```diff
@@ -76,7 +77,28 @@ public void StartNextRound()
...
+
+        // TODO assign chips
+
+        CheckWinCondition();
+    }
+
+    private void CheckWinCondition()
+    {
+        var lastWinner = rounds.Last().winner;
+        var wins = rounds.Select(round => round.winner == lastWinner).Count();
```
Fueredoriku, line 112: This seems to always count every round won no matter which player was last winner, which makes the player who wins the third round always be the match-winner. 🤔

Assets/Scripts/Control&Input/PlayerMovement.cs
```diff
@@ -11,6 +11,8 @@ enum PlayerState
     DEAD
 }
 
+[RequireComponent(typeof(Rigidbody))]
+[RequireComponent(typeof(Collider))]
```
Stektpotet, line 15: `RequireComponent` allows for up to three components to be required:

```cs
[RequireComponent(typeof(Rigidbody), typeof(Collider))]
public class PlayerMovement : MonoBehaviour {...}
```

toberge: Thanks, refactored as per your suggestion.

Assets/Scripts/Gamestate/MatchController.cs
```diff
@@ -0,0 +1,141 @@
...
+    public MatchPlayer(Player player, PlayerStateController playerStateController, int startAmount)
+    {
+        this.player = player;
+        playerStateController.chips = startAmount;
+        this.playerStateController = playerStateController;
+    }
+    // Reference to player identity class
+    public Player player;
+    // Reference to in-match player
+    public PlayerStateController playerStateController;
```
Stektpotet, line 22: The comment and my  _professional_ opinion on variable and type naming make begs the question if `PlayerStateController` cannot be renamed to something more explicit than `Controller`

Assets/Scripts/Gamestate/Round.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using UnityEngine;
+
+public class Round
+{
+    public PlayerStateController winner;
+    public Dictionary<PlayerStateController, List<PlayerStateController>> kills = new Dictionary<PlayerStateController, List<PlayerStateController>>();
+    public List<PlayerStateController> players = new List<PlayerStateController>();
+    public List<PlayerStateController> livingPlayers = new List<PlayerStateController>();
```
Stektpotet, line 11: public access 🤔🤔 - a recipe for potential future bugs!


In this case I'm all for privatisation!

Some of the fields may yet need read access externally, so maybe we can add some properties?
Getter property for `winner`, and maybe `IReadOnly...`-properties wrapping the dictionaries and lists for external access.
Otherwise, I think these in general should only be modifiable internally (i.e. private access), and through member method calls to the round itself.

Assets/Scripts/Gamestate/Round.cs
```diff
@@ -0,0 +1,61 @@
...
+
+    public void OnRoundEnd()
+    {
+        foreach (var player in players)
+        {
+            player.onDeath -= OnDeath;
+        }
+        MatchController.Singleton.onRoundEnd -= OnRoundEnd;
+    }
+
```
Stektpotet, line 32: Add something like this for ease of access to stats?

```cs
// Round.cs
public int KillCount(PlayerStateController player)
{
    if (kills.TryGetValue(player, out var playerKills))
        return playerKills.Count;
    return 0;
}

public bool IsWinner(PlayerStateController player)
{
    return player == winner;
}
```

Assets/Scripts/Gamestate/MatchController.cs
```diff
@@ -0,0 +1,141 @@
...
+            player.playerStateController.chips +=
+                // Base reward
+                rewardBase
+                // Kill bonus
+                + (rounds.Last().kills.ContainsKey(player.playerStateController) ?
+                rounds.Last().kills[player.playerStateController].Count * rewardKill : 0)
+                // Winner bonus
+                + (player.playerStateController == rounds.Last().winner ? rewardWin : 0);
+        }
+    }
```
Stektpotet, line 108: Although I also tend to (mis)use the ternary operator, I find that in this scenario, we might be better off with more expressive if-statements, or even better: methods defining these stats in the `Round`-class (see comment on Round.cs)

e.g:
```cs
// MatchController.cs
private void AssignRewards()
{
    var lastRound = round.Last();
    foreach (MatchPlayer player in matchPlayers)
    {
        var reward = rewardBase + lastRound.KillCount(player.playerStateController) * rewardKill;
        if (lastRound.IsWinner(player.playerStateController)
           reward += rewardWin;

        player.playerStateController.chips += reward;
    }
}
```

Stektpotet (changes requested):
Overall clean, expressive code, and has okay readability.

I've added some nitpicking comments, though my biggest concerns regard the naming of things, and accessor modifiers.

Overall, great work! 🥳 🎮👾

Assets/Scripts/Gamestate/Round.cs
toberge: I agree that these fields might be better off private, and your suggestions seem well-thought-out.
If I remember correctly, this class was intended to be a simple storage class, @Fueredoriku had some thoughts on its responsibilities that he might want to mention here. Implementing your suggestions should make the code using this class a little cleaner in any case.

Assets/Scripts/Gamestate/MatchController.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        PlayerInputManagerController.Singleton.playerInputs.ForEach(playerInput =>
+        {
+            var player = playerInput.gameObject.GetComponent<Player>();
+            var playerStateController = playerInput.gameObject.transform.parent.GetComponent<PlayerStateController>();
+            matchPlayers.Add(new MatchPlayer(player, playerStateController, startAmount));
+        });
+
+        // TODO do something else funky wunky
+        onRoundEnd += () => Debug.Log("End of round " + rounds.Count());
```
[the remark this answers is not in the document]
toberge: Leaving this line as it is, should not create any _real_ problems, and it is only intended as a way for us to know the round number before we implement any UI for it.

Assets/Scripts/Gamestate/MatchController.cs
toberge: Agreed, though I'm not sure what we should rename it to. Simply `PlayerState`?

Assets/Scripts/Gamestate/MatchController.cs
```diff
@@ -0,0 +1,141 @@
...
+                Debug.LogWarning($"There's more than one {Singleton.GetType()} in the scene!");
+                Destroy(gameObject);
+            }
+
+            return;
+        }
+
+        Singleton = this;
+
+        #endregion Singleton boilerplate
```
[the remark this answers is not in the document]
Fueredoriku: Do we want to resolve the Singleton boilerplater restructure in this PR or handle it as a seperate Issue in #30 ? I believe there are other instances of this boilerplate code not present in this PR too 🤔

Assets/Scripts/Gamestate/MatchController.cs
Fueredoriku: What we want to describe with the name is "All information a single player holds exclusive to a single match", of which I had hard time to create a good name from. Very open for suggestions 😅

Assets/Scripts/Gamestate/Round.cs
Fueredoriku: IReadOnly can be a nice solution.

I do not want to make all these fields private as there are potentially classes 1 or more steps away from MatchManager that can be interested in things like round statistics (kills, winners, etc.), which would make acessing these fields a pain.

During a round we would need to store a private "writeable" collection (as it keeps track of the round dynamically), and then save this as public ReadOnly collection after the round has concluded and becomes a "storage only"-object.
It would require some refactoring (and a bit more code) in the Round class, but its negligble enough that we can do it to prevent abuse of the collection fields I guess 😅. (I'm atleast a fan of restricting architecture if you want to underline that something might be weird with your approach if you need unreachable access)

Assets/Scripts/Gamestate/MatchController.cs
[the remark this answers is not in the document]
toberge: I'd agree with doing it in a separate PR.

Assets/Scripts/Gamestate/MatchController.cs
TArchetype: As per discussed in meeting, class is renamed to simply `Player`

toberge: No, in fact it is renamed to `PlayerManager`, `Player` is renamed to `PlayerIdentity` and `MatchPlayer` is renamed to `Player`. ~~This makes sense, I swear.~~

Stektpotet (approved):
Okay! Go ahead, I might have a look at persistent readonly backingfields instead of creating new ones upon every getter-property call in future for the Rounds, but this is good for now!

PR #23, opened 2022-11-07
Feature/basic gunplay

SondreHus:
Have fun, suckers


Assets/Editor/GunControllerEditor.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        DrawDefaultInspector();
+
+        GunController myScript = (GunController)target;
+        if (GUILayout.Button("Build Gun"))
+        {
+            myScript.InitializeGun();
+        }
+    }
+}
```
toberge, line 18: Missing newline at end of file :anger:

Assets/Editor/ModifyableFloatDrawer.cs
```diff
@@ -0,0 +1,13 @@
+using UnityEditor;
+using UnityEditor.UIElements;
+using UnityEngine;
+
+[CustomPropertyDrawer(typeof(ModifyableFloat))]
+public class ModifyableFloatDrawer : PropertyDrawer
```
toberge, line 6: Modif*i*able with an *i*, my guy

Assets/Editor/GunControllerEditor.cs
```diff
@@ -0,0 +1,18 @@
+using UnityEngine;
+using System.Collections;
+using UnityEditor;
+
+[CustomEditor(typeof(GunController))]
+public class GuncControllerEditor : Editor
```
Stektpotet, line 6: Typo?

Assets/Editor/GunControllerEditor.cs
Stektpotet, line 18: Missing newline

Assets/Scripts/GunController.cs
```diff
@@ -0,0 +1,134 @@
...
+        // Destroys gun child before construction
+        for (int i = this.transform.childCount; i > 0; --i)
+            DestroyImmediate(this.transform.GetChild(0).gameObject);
+
+        // Instantiates the different parts
+        var gunBody = Instantiate(bodyPrefab, transform)
+            .GetComponent<GunBody>();
+
+        // Stats is retrieved from gun body
+        // NEVER REFERENCE THE GUNSTAT PREFAB DIRECTLY, BUNBODY AUTOMATICALLY INSTANTIATES IT
```
toberge, line 55: bunbody <3

Assets/Scripts/GunModifyer.cs
```diff
@@ -0,0 +1,62 @@
...
+
+[System.Serializable]
+public class Modifyer
+{
+    public string name;
+    public float addition = 0;
+    public float multiplier = 0;
+    public float exponential = 1;
+}
+public class GunModifyer : MonoBehaviour
```
toberge, line 14: modifier...

Assets/Scripts/GunModifyer.cs
```diff
@@ -0,0 +1,62 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEditor;
+
+[System.Serializable]
+public class Modifyer
```
toberge, line 7: modifier...

Assets/Scripts/GunModifyer.cs
```diff
@@ -0,0 +1,62 @@
...
+
+    // Used to keep track of added projectile modfyer to delete it 
+    private GameObject instantiatedBulletModifyer;
+
+    // Is run when the gun is built
+    public virtual void Attach(GunController gun)
+    {
+        foreach(var modifier in statModifiers)
+        {
+            //print(((ModifyableFloat)gun.stats.GetType().GetProperty("Firerate", typeof(ModifyableFloat)).GetValue(gun.stats, null)).value());
```
toberge, line 34: Remove or keep as Debug.Log

Assets/Scripts/GunModifyer.cs
```diff
@@ -0,0 +1,62 @@
...
+            stat.addBaseValue(modifier.addition);
+            stat.addMultiplier(modifier.multiplier);
+            stat.addExponential(modifier.exponential);
+        }
+        if(bulletModifyerPrefab != null)
+        {
+            instantiatedBulletModifyer = Instantiate(bulletModifyerPrefab, gun.projectile.transform);
+        }
+    }
+    // Is run when component is remove
```
toberge, line 46: removed, also whitespace :))))

Assets/Scripts/GunController.cs
```diff
@@ -0,0 +1,134 @@
...
+
+    // All the stats of the gun and projectile
+    public GunStats stats{ get; private set; }
+
+    // Inputs
+    public bool triggerHeld, triggerPressed;
+
+
+    //Projectile to shoot
+    [HideInInspector]
```
Stektpotet, line 34: Attributes unused by properties, probably a reminder from a time past when `projectile` was a field?

Assets/Scripts/GunStats.cs
```diff
@@ -0,0 +1,93 @@
...
+
+[CreateAssetMenu(fileName = "GunStats", menuName = "ScriptableObjects/GunStats", order = 1)]
+public class GunStats: ScriptableObject
+{
+
+    // -- IMPORTANT INFO --
+
+    // It is fine to add new properties to this file, but be carefull when REMOVING them
+    // If you remove, for example, magazineSize from these properties and then save, all object data of magazine size 
+    // will be removed in every instance of the sriptable object
```
toberge, line 11: Put as `<summary>` in `///` comment, perhaps?

Assets/Scripts/GunStats.cs
```diff
@@ -0,0 +1,93 @@
...
+    public ModifyableFloat ProjectileScale { get => projectileScale; }
+
+
+    // How much extra damage a crit does, the standard is a crid does double damage
+    [SerializeField]
+    private ModifyableFloat criticalMultiplier = new ModifyableFloat(2f);
+    public ModifyableFloat CriticalMultiplier { get => criticalMultiplier; }
+
+
+    // TODO: make modifyableInteger
```
toberge, line 87: modifiable mwahahahaha

Assets/Scripts/HealthController.cs
```diff
@@ -0,0 +1,34 @@
...
+    public int CurrentHealth { get => currentHealth; }
+
+    public delegate void damageEvent(HealthController healthController, int damage);
+    public damageEvent OnDamageTaken;
+
+    private void Awake()
+    {
+        foreach(var box in GetComponentsInChildren<HitboxController>())
+        {
+            box.health = this;
```
toberge, line 20: Yeah we'll deal with the structure of this later. Not sure if this is ideal :)))))

Assets/Scripts/ProjectileController.cs
```diff
@@ -0,0 +1,95 @@
...
+        return Vector3.Lerp(state.oldPosition, state.position, diff);
+    }
+
+    //public Vector3 qerpPos()
+    //{
+    //    var diff = (Time.time - state.lastUpdateTime) / Time.fixedDeltaTime;
+    //
+    //    return ((state.olderPosition + state.position) * 0.5f - state.oldPosition) * diff * diff +
+    //        (state.position - state.olderPosition) * diff + state.oldPosition;
+    //}
```
toberge, line 94: Why is this commented out?

Assets/Scripts/SpiralPathModifier.cs
```diff
@@ -0,0 +1,44 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class SpiralPathModifier : MonoBehaviour
```
toberge, line 5: ok :)

Assets/Scripts/Utils/ModifyableFloat.cs
```diff
@@ -0,0 +1,47 @@
...
+    private float multiplier = 1f;
+
+    private float exponential = 1f;
+
+    public  ModifyableFloat(float value)
+    {
+        baseValue = value;
+    }
+
+    public void addMultiplier(float value)
```
toberge, line 22: i'm gonna add a muliplyer >:)

toberge (changes requested):
Nitpicks w/o testing

Assets/Scripts/ProjectileController.cs
SondreHus: It behaved poorly, but ought to be mathematically correct, if we hapen to be interested in trying a quadratic interpolation later.

Fueredoriku (changes requested):
Very cool feature! 😁

This system is very complex, so I had to map it out to properly give feedback related to the architecture and logic. In addition I have some feedback regarding the semantics of some scripts, which I left as comments.

Below is a class diagram of the augment system, where I've left my remarks as comments with yellow backgrounds:
AugmentSystem drawio

Here are the main points:
- I think GunController should be seperated into a "GunFactory" class that holds all the gun initialization logic and a "GunController" that holds all FireGun-related logic.
  - The first argument for doing this is for structure. Keeping instantiation logic and functional shooting logic  separate would adhere to the single responsibility principle, as they can (and in my opinion) should function separetly, as a GunController initialized by a GunFactory could set all dependabilities in it's constructor.
  - The second argument is for perfomance. Having the entire gun initialization logic attached to each player's weapon at all times seems redundant when all you need is the shooting-related logic. In theory there is only need for one gunFactory that constructs all weapons.
  - The third argument is that the PlayerStateManager from PR #22 (and the rest of the structures) would have a better time accessing relevant gun info and ONLY relevant gun info from GunController without going through other classes they don't really care about.
- I think FireRateController should be an interface implemented by the different FireRateController-classes.
  - It seems a bit weird to me to have an inner abstract class FireRateController instread of a interface. Different new FireRate classes seems like something we would like to implement in the future, and only having to extend a interface is better for modifyability in such a case. 
- I think the GunBody-class should extend GunModifier, as it currently isn't attached to any of it's sibling classes from a logical view perspective.
  - I'm assuming GunBody doesn't extend GunModifier because it differentiates it from it's siblings by providing the "base" GunStats that for the other sibling components. I think this should rather be solved by having a GunFactory create the "base" GunStats, and send that to the GunBody which would extend GunModifier, thus working more in line with the other sibling components.
- SpiralPathModifier and future scripts like it intended for being components of the bulletprefab (I assume that's how they are supposed to be used), should probably have an interface or alternatively (but porbably a bit weird) extend the modifier class.

Also optional things:
- As mentioned in our previous meeting, functionality for applying effects to the shooter when firing could be nice, but that something we could always implement later.
- I don't quite see the reasoning for having a private collection of HealthControllers in ProjectileDamageController. If you want persistent damage effects over time and need them for this case, then it would probably be better to set that to PlayerStateManger instead. (This feature is however still up for review, which is why can leave this as is)

Assets/Scripts/FireRateController.cs
```diff
@@ -0,0 +1,69 @@
+using UnityEngine;
+
+public abstract class FireRateController
+{
+    public abstract bool shouldFire(bool triggerPressed, bool triggerHeld);
```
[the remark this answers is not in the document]
SondreHus: Yes, for instance a chanrgin weapon would not need the same field, but rather (for inctance) chargeRate and currentCharge.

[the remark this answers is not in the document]
Fueredoriku: In that case this suggestion can be ignored :)

SondreHus:
> The second argument is for perfomance. Having the entire gun initialization logic attached to each player's weapon at all times seems redundant when all you need is the shooting-related logic. In theory there is only need for one gunFactory that constructs all weapons.

The performance impact of having them combined is completely negligible, also they should be available

Assets/Scripts/ProjectileController.cs
toberge: Perhaps we could remove the `//`es and leave it as an unused method?

Assets/Scripts/BulletController.cs
```diff
@@ -0,0 +1,70 @@
...
+        foreach(var collision in collisions)
+        {
+
+            OnColliderHit?.Invoke(collision, ref state, stats);
+            HitboxController hitbox = collision.GetComponent<HitboxController>();
+
+            if (hitbox != null){
+                OnHitboxCollision?.Invoke(hitbox, ref state, stats);
+            }
+        }
```
Stektpotet, line 61: As is, the colliders don't need to be stored in an array. You can simply pass them on as an `IEnumerable<Collider>` out of the `Select`-call:

I would also argue that the hit info could be interesting to pass on to the various events, in case the events could use them for further effects... 🤔 

i.e:
```cs

public delegate void CollisionEvent(RaycastHit hit, ref ProjectileState state, GunStats stats);

//...

foreach(var raycastHit in rayCasts.OrderBy(x => x.distance))
{
    OnColliderHit?.Invoke(raycastHit, ref state, stats);
    HitboxController hitbox = raycastHit.collider.GetComponent<HitboxController>();

    if (hitbox != null)
    {
        OnHitboxCollision?.Invoke(hitbox, ref state, stats);
    }
}
```

Assets/Scripts/GunBody.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+
+public class GunBody : MonoBehaviour
+{
+    [SerializeField]
+    private GunStats stats;
+
+    // Base stats of the gun
+    public GunStats Stats { get => Instantiate(stats); }
```
Stektpotet, line 11: Maybe I've misunderstood something here, but I think It'd be ideal to only instantiate it once?

```cs
// template scriptableobject
[SerializeField]
private GunStats stats;

// Base stats of the gun
private GunStats statsInstance;
public GunStats Stats 
{ 
    get 
    {
        if (statsInstance == null)
            statsInstance = Instantiate(stats); 
        return statsInstance;
    }
}

File: Assets/Scripts/GunStats.cs [hunk 19]
```diff
@@ -0,0 +1,93 @@
...
+        fullAuto
+    }
+    // Projectiles per second
+    [SerializeField]
+    private ModifyableFloat firerate = new ModifyableFloat(5f);
+    public ModifyableFloat Firerate { get => firerate;}
+
+    // Time to reload gun
+    [SerializeField]
+    private ModifyableFloat reloadTime = new ModifyableFloat(3f);
```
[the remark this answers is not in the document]
Stektpotet: Actually, I've seen the naming convention for properties be UpperCamelCase. Examples of this can be found all over in the standard packages. I think Unity is the one deviating here...

image
image

Stektpotet (changes requested):
MUST FIX:
Missing brackets in `ModifyableFloatDrawer.cs`

Other than that, the comments I've added should be sufficient.

Assets/Scripts/GunStats.cs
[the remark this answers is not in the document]
Fueredoriku: Interesting 🤔 We should probably decide on which to follow in that case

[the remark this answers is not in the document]
toberge: Hm, this spices things up :upside_down_face: 
I assumed Unity's naming convention was in line with C#'s and didn't think of the properties that happen to be PascalCase (though they've thrown me for a loop when I've assumed that e.g. array.Count is a method)

toberge:
> > The second argument is for perfomance. Having the entire gun initialization logic attached to each player's weapon at all times seems redundant when all you need is the shooting-related logic. In theory there is only need for one gunFactory that constructs all weapons.
> 
> The performance impact of having them combined is completely negligible, also they should be available

Performance shouldn't be an issue here, the instruction won't be duplicated for each instance or anything like that.
What I personally find most aggravating about the architecture of the gun stat system is that the base does not inherit from the modifier class.

As for the comments regarding the hitbox collisions, I suggest we leave changes to that for a future pull request where we actually integrate it with the players. Just to document my thoughts: I would rewrite it to invoke OnHit delegates on the colliders that are hit, and let the objects that are hit handle what happens to them without storing information about which colliders were hit, as that seems redundant.

Assets/Scripts/GunBody.cs
Stektpotet: Or, if not: I think it should be explicitised that we're making a new instance. For example by changing the getter-property to a function aptly named to clarify the instantiation.

SondreHus: The gunbody Stats will usually only be called once and passed as an instance to guncontroller. Whenever we remake the gun, it will be called again, reseting the stats, which is intentional. However this could be trouble is someone decides to get the stat directly from the body.

Assets/Scripts/BulletController.cs
SondreHus: Not all oncolliderhits will use raycast, some will probably use hitboxes and rigidbodies, if for instance we want grenades

Assets/Editor/GunControllerEditor.cs
toberge: There's still a missing newline at the end of this file? :triumph:

Assets/Scripts/GunModifyer.cs
```diff
@@ -4,21 +4,21 @@
...
-public class Modifyer
+public class Modifier
 {
     public string name;
     public float addition = 0;
     public float multiplier = 0;
     public float exponential = 1;
 }
-public class GunModifyer : MonoBehaviour
+public class GunModifier : MonoBehaviour
```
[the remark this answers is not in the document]
toberge: This ain't done yet?

Assets/Scripts/GunStats.cs
```diff
@@ -68,26 +68,26 @@ public enum FireModes
...
+    private ModifiableFloat projectileScale = new ModifiableFloat(1f);
+    public ModifiableFloat ProjectileScale { get => projectileScale; }
 
 
     // How much extra damage a crit does, the standard is a crid does double damage
     [SerializeField]
-    private ModifyableFloat criticalMultiplier = new ModifyableFloat(2f);
-    public ModifyableFloat CriticalMultiplier { get => criticalMultiplier; }
+    private ModifiableFloat criticalMultiplier = new ModifiableFloat(2f);
+    public ModifiableFloat CriticalMultiplier { get => criticalMultiplier; }
```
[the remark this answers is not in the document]
toberge: should be fine for now :slightly_smiling_face:

Assets/Editor/GunControllerEditor.cs
toberge: eh, it's fiiiiiiiiiiiiiine

toberge (approved):
it fine, Sondre. There are only petty nitpicks left, let's just get this over with :)))))

Fueredoriku (approved):
This works for me! 😎

Assets/Scripts/GunBody.cs
toberge: I'm slightly concerned, but this should be fine as long as we don't use it wrongly

Stektpotet (approved):

PR #36, opened 2022-11-15
Physical bidding round 

Fueredoriku:
TODO:
- Create a physical bidding round scene
- Integrate bidding round to MatchManager


Assets/Scripts/Gamestate/PlayerFactory.cs
```diff
@@ -27,21 +28,35 @@ private void Awake()
     /// </summary>
     private void TransferExistingInputs()
     {
+        if(playerInputManagerController == null) { Debug.Log("Go to current scene from Menu to actually spawn players"); }
+        //TODO: (not this, preferrably make playerfactory intstantiate by call instead of awake)
+        if (SceneManager.GetActiveScene().name == "Bidding")
```
Fueredoriku, line 33: TODO: Not this ugly check

Fueredoriku: I'll fix this for the bidding -> round implemntation later on

Assets/Scripts/Gamestate/MatchController.cs
```diff
@@ -83,13 +83,31 @@ public void StartNextRound()
...
+    }
+
     public void EndActiveRound()
     {
         onRoundEnd?.Invoke();
 
         AssignRewards();
 
-        CheckWinCondition();
+        if (!CheckWinCondition())
```
toberge, line 99: Perhaps we should rename this method now that it returns a value?

Assets/Scripts/Gamestate/PlayerManager.cs
```diff
@@ -17,11 +17,20 @@ public class PlayerManager : MonoBehaviour
...
     [SerializeField]
     private GameObject meshBase;
 
+    [SerializeField]
+    private List<Item> items;
+
+    public void PerformTransaction(Item item, int cost)
+    {
+        items.Add(item);
+        chips -= cost;
```
toberge, line 31: Perhaps we should perform *some* sanity check here, instead of naively subtracting the cost from our chip count?

Assets/Scripts/Gamestate/PodiumController.cs
```diff
@@ -0,0 +1,46 @@
...
+    void AddBid(InputAction.CallbackContext ctx)
+    {
+
+    }
+
+    // Update is called once per frame
+    void Update()
+    {
+        
+    }
```
toberge, line 45: Should this be empty? Should this method be implemented here at all?

Assets/Scripts/Gamestate/PodiumController.cs
```diff
@@ -0,0 +1,46 @@
...
+    private MatchController matchController = MatchController.Singleton;
+
+    void Start()
+    {
+    
+    }
+
+    private void OnTriggerEnter(Collider other)
+    {
+        PlayerManager player = GetComponentInChildren<PlayerManager>();
```
toberge, line 22: **??????**

How does this PodiumController have the specific player it collides with as a child?

toberge (changes requested):
LGTM for a temporary, nonfunctional version of the bidding scene. I've left a few comments.

Assets/Scripts/Gamestate/MatchController.cs
Fueredoriku: Yes

Assets/Scripts/Gamestate/PodiumController.cs
Fueredoriku: Podium is kill

Assets/Scripts/Gamestate/PlayerManager.cs
Fueredoriku: We might need to consult @Stektpotet regarding this, as it's only imported code.

I do however think this code is subject to change in #44 , so I'll await this I think

toberge: Sounds like we should put this off, yeah.

toberge (approved):

PR #39, opened 2022-11-16
Fix/round readonly properties

Stektpotet:
As mentioned in #22, I wanted to modify how the initialisation of the `ReadOnly...` members happened. This does exactly that:

Change Round initialisation to make fields `readonly` and to make collections' contents also remain read-only when accessed externally.


Fueredoriku (approved):
Elegant solution! Haven't gotten the chance to test the code, but it seems sound, so I'll give it the "👍"

PR #49, opened 2022-11-24
Feature/add basic gun parts

toberge:


toberge:
current issue: revolver base gets placed too far back :(

image

toberge:
> current issue: revolver base gets placed too far back :(
> 
> image

Possible solution: Let us say that each gun body has origin where the hand of the player model would be. This makes no sense at the moment, given our sussy player models (we need some **hand**some fellas), but should give us a unified starting point for smacking guns onto our amogii so it doesn't look like the crap it does right now.

toberge:
So each gun base should have origin like this:

image

Assets/Scripts/GunExtension.cs
```diff
@@ -14,12 +14,27 @@ public class GunExtension : GunModifier
...
+        var attachedOutputs = new List<Transform>();
+        foreach (var t in transforms[1..])
         {
             Instantiate(model, t.position, t.rotation, transform);
+            foreach (var output in outputs)
+            {
+                // TODO Do we need a rotational offset?
+                var offset = output.position - transforms[0].position;
+                var outputInstance = Instantiate(output, t.position + offset, t.rotation, transform);
+                attachedOutputs.Add(outputInstance.transform);
```
toberge, line 35: Resolved Sondre's complaint about extension instantiation, though I've left a TODO that is irrelevant right now. We might need to change this bit for future spicy rotations in extension outputs :)

toberge: Made an attempt that did not work properly:

```cs
var position = t.position + (output.position - transforms[0].position);
var rotation = (output.rotation * Quaternion.Inverse(transforms[0].rotation)) * t.rotation;
var outputInstance = Instantiate(output, position, rotation, transform);
```

Fueredoriku: Can be solved later, I would say it's ok to leave the TODO.

Fueredoriku (approved):
Looks good! Read code and tested 👍

PR #61, opened 2022-11-30
A shared HUD with a round timer

Fueredoriku:
Includes both the visual and functional parts of the round timer.

Things to note:
- If a round timer runs out, players are sent to bidding with no winner for the round.
- Timer position on screen is dependent on number of players to ensure best visuals no matter the player count.


Assets/Scripts/Gamestate/MatchController.cs
```diff
@@ -43,6 +46,11 @@ public class MatchController : MonoBehaviour
     [SerializeField]
     private int rewardBase = 1;
 
+    public Timer roundTimer;
+
+    [SerializeField]
+    private HUDGlobalController globalHud;
```
toberge, line 52: GlobalHUDController and globalHUD?

toberge (commented):
Just a naming suggestion, I haven't tested this (though it probably runs fine).

Assets/Scripts/Gamestate/MatchController.cs
Fueredoriku: As we have have a HUDController, I wanted to build on that specific name for something functionally similar. But I don't really have a strong opinion regarding this, so I can change it before merge

toberge: I think we should rename the HUDController to something more specific though, like PlayerHUDController or LocalHUDController

Assets/Scripts/Gamestate/MatchController.cs
```diff
@@ -43,6 +46,11 @@ public class MatchController : MonoBehaviour
     [SerializeField]
     private int rewardBase = 1;
 
+    public Timer roundTimer;
+
+    [SerializeField]
+    private GlobalHUDController globalHUDController;
```
TArchetype, line 52: Since this component is on the same game object  consider adding a `[requireComponent()]` field and so referencing that through a `GetComponent()` call.

Assets/Scripts/Timers/Timer.cs
```diff
@@ -38,6 +38,10 @@ public class Timer : MonoBehaviour
     /// Runs once when the timer stops.
```
TArchetype, line 38: Small nitpick, but you can remove `System.Collections.Generic` and `System.Data` namespaces from this script.

Assets/Scripts/UI/GlobalHUDController.cs
```diff
@@ -0,0 +1,30 @@
...
+
+    void Start()
+    {
+        if (PlayerInputManagerController.Singleton.playerInputs.Count > 2)
+        {
+            roundTimer.GetComponent<RectTransform>().anchoredPosition = Vector2.zero;
+        }
+        else
+        {
+            roundTimer.GetComponent<RectTransform>().anchoredPosition = new Vector2(0, textOffsetY);
```
TArchetype, line 21: The timer text is not positioned correctly for when we have one or two players:

image
image

Consider expanding the `Text  (TMP)` component to cover the entire screen, then instead of setting the anchored position, switch between top centered text and middle centered text options in the text itself.

Assets/Scripts/Timers/Timer.cs
```diff
@@ -38,6 +38,10 @@ public class Timer : MonoBehaviour
...
     /// </summary>
```
TArchetype, line 39: Is there any need to Serialize the timer components if they are all controlled by a different script?

TArchetype (changes requested):
This is a great feature to have, adds a certain level of pressure to the players and keeps things interesting. It would also be great in the future to expand on this feature by adding an end of round result page which visually shows how much resources players earned in this match and have.

However I do have a couple of comments before merging:

Assets/Scripts/Timers/Timer.cs
Fueredoriku: You should discuss that with the author @Stektpotet, as it's outside the scope of the changes I implemented in this PR

Assets/Scripts/Gamestate/MatchController.cs
Fueredoriku: I disagree for multiple reasons:

- Since this is in the scope of MatchController, this would have to be a getComponentInchildren or have a refference to the HUD object child which defeat the purpose.
- The matchmanager might find itself in situations where it doesn't even need the global HUD (like being in the bidding-scene).
- GetComponent is nice to avoid when it makes sense. MatchController itself being a required component on the MatchManager itself already implies that this script is prefab specific, meaning people are unlikely to find the requiredComponent useful.

Assets/Scripts/UI/GlobalHUDController.cs
Fueredoriku: The text is displayed correctly with any amount of players as long as the aspect ratio is normal. I will make a commit to make it work with other aspect rations than 16:9 though.

toberge (changes requested):
LGTM, as long as you fix the aspect ratio issue @TArchetype mentioned and also the brief moment where the timer is **-1:0-1** before the bidding scene loads, it's ready to merge.

toberge (approved):
Given that we will replace the timer with some other UI when the time is up, the -1:-01 issue is insignificant.

TArchetype (approved):
Timer position fixed, approved

PR #63, opened 2022-11-30
SetPlayerInput functions now uses inputManagers instead

Fueredoriku:
- InputManager already has accessible field for playerInput
- Now getComponent is only called once when a new player join (instead of the previous 10!!)


toberge (approved):
Looks like a good change to me. Approving this, though I'm not sure if we'll run into issues with the options menu PR.

PR #64, opened 2022-12-01
gunstatsHUD for bidding scene

Fueredoriku:
A gunstats and general info HUD that display properties like name and description for augments up for auction viewable in bidding scene.


Fueredoriku:
image

Fueredoriku:
After the last meeting, it was decided that a new design of the bidding UI should be used, this PR should thus be closed.

PR #80, opened 2023-02-08
Add gun stat UI to bidding scene

toberge:
# Functional

- [x] Player name and chip count
- [x] Stat bars for gun stats with dynamic delta color
- ~~Dynamic set of bars depending on the most impactful change~~

# Visual

- [x] Interpolate transitions
- [ ] Pulsing deltas (not important)

# Current status


TArchetype (approved):

Fueredoriku (approved):
Looks very nice! 😄👍

The only feedback I have is that when changing from bidding scene back to round, the visual gunstats show some weird deltas. THis could be fixed with a check in updateMeter or something, but that isn't too important before we actually implement the concrete visual implementation of the UI anyways.

Code read and tested, so I approve! 😁

PR #81, opened 2023-02-09
Feature/bidding models

TArchetype:
**Summary:**
Includes a new bidding scene, some demo comparison UI elements and minor assets such as tables, chairs and light-fixtures.

image


Fueredoriku (approved):
Looks very nice! 🎉

Added a mesh collider so paluyer's don't fall trhough, but I apporve of the current state 👍

PR #89, opened 2023-02-20
Round ends quickly if in "testing mode"

Fueredoriku:
Testing mode entails joining a match alone with the unity editor instead of a built version of the game.

If this is the case, the shooting round will end in 5 seconds.


toberge (approved):
LGTM

PR #93, opened 2023-02-21
Integrate player stats with bidding scene

toberge:
# Purpose
Improve UI y'know

# What's done already
- [x] Fix bidding platforms not being wired up to controller
- [x] Assign player stat UIs to panels
- [x] Adjust UI to look okay
- [x] Limit scaling of stats to always adjust for a new max
- [x] Put labels above their stat bar
- [x] Hide stats for missing players ;)

# Future improvements
- [ ] Add icons (text looks bad)

### DIFF
--- a/Assets/Prefabs/UI/PlayerStatUI.prefab
+++ b/Assets/Prefabs/UI/PlayerStatUI.prefab
@@ -9,10 +9,10 @@ GameObject:
   serializedVersion: 6
   m_Component:
   - component: {fileID: 1754669797276723109}
-  - component: {fileID: 2909222001333712944}
-  - component: {fileID: 5514397517174202729}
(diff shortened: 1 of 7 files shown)


Fueredoriku (approved):
Looks great! 👍

Read code and tested, works fine on my end.

As I mentioned we might want to think about "bakeing" some state together (eg form DPS), but I'll create a separate issue for that I think.

Great work, I approve! :)

toberge:
> Looks great! 👍
> 
> Read code and tested, works fine on my end.
> 
> As I mentioned we might want to think about "bakeing" some state together (eg form DPS), but I'll create a separate issue for that I think.
> 
> Great work, I approve! :)

Yeah, the stats aren't final here, I just put together the most reasonable set of stats among the attributes in GunStats.

PR #107, opened 2023-03-07
Fix duplicate gun spawning on InitializeGun

### LINKED ISSUE #91: Gun in bidding scene
Gun is currently bugged by being carried over from shooting scene.

This could be a nice feature if implemented properly, where you can test the gun with the new parts at the biddingplatforms.

toberge:
GunFactory is a MonoBehaviour and is thus spawned in when one of its
static methods is called. In its Start method, it calls InitializeGun
to instantiate the gun it is supposed to build. This is *only* useful
in the GunTest scene, where we want a gun to show up automatically,
based on what the factory starts with. When we use InstantiateGun to
create a gun from other scenes, the Start method is also triggered and
creates a superfluous copy of the gun.

With this commit, the issue is avoided by initializing in InstantiateGun
and in Start *only when the scene in GunTest*.

Fixes #91

### DIFF
--- a/Assets/Scripts/Augment/GunFactory.cs
+++ b/Assets/Scripts/Augment/GunFactory.cs
@@ -1,6 +1,7 @@
 using System.Collections.Generic;
-using UnityEngine;
 using System.Linq;
+using UnityEngine;
+using UnityEngine.SceneManagement;
 
 [RequireComponent(typeof(GunController))]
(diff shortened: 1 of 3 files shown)


Fueredoriku (approved):
Looks good! 😎👍

Made a small commit to update gun position in bidding, other that that I have no comment

I approve!

PR #110, opened 2023-03-08
Add leading bidder indicator to bidding platforms

### LINKED ISSUE #73: Indication of which player has the leading bid on an item

toberge:
- [x] Replace bidding platform model with one with more suitable UV mapping
- [x] Add material that displays a border in the leading bidder's color

Resolves #73.

image

And since it's related to the bidding scene, I fixed the padding of PlayerStatUI :cry:


Fueredoriku (approved):
A great and needed feature! 😎👍

Code looks code and runs as expected,
good work!

PR #113, opened 2023-03-15
Solar body augment

Fueredoriku:
A solar body augment that fills gun magazine when players face the correct angle
- Added reloading mechanic

More polish is wanted, but will be postponed for later

Updating the weapon transform in fps mode is going to be separate PR.


Assets/Scripts/Augment/AugmentImplementations/Hat/HatBarrelModel.cs
```diff
@@ -19,42 +19,36 @@ public class HatBarrelModel : MonoBehaviour
...
-    void Start()
+    public void OnInitialize(int magazine)
     {
-        remainingBullets = magazineSize;
+        magazineSize = magazine;
         for (int i = 0; i < magazineSize; i++)
         {
             ammunition.Add(Instantiate(hatPrefab, ammunitionHolder.transform));
             ammunition[i].transform.Translate(new Vector3(0f, bulletHeigth * i, 0f));
         }
```
toberge, line 30: This whole class feels like something that could work for most barrels. Perhaps we could refactor it (later)?

Assets/Scripts/Augment/GunBody.cs
```diff
@@ -12,4 +12,33 @@ public class GunBody : MonoBehaviour
 
     // Where to attach barrel
     public Transform attachmentSite;
+
+    [SerializeField, Range(0,1)]
+    protected float reloadEfficiencyPercentage = 1f;
+
+    //TODO: Modifier refactor
+    protected GunController gunController;
```
toberge, line 20: Elaborate? I assume you mean that we should give access to the gun controller in some other way?

Assets/Scripts/Augment/GunController.cs
```diff
@@ -36,8 +37,23 @@ private void FixedUpdate()
         }
     }
 
+    /// <summary>
+    /// Expects a percentage of ammunition to be reloaded.
+    /// This percentage is normalized eg. min = 0, max = 1.
```
toberge, line 42: Ummmm so it's just a fraction. No need to call it a percentage when it isn't one.

toberge (changes requested):
Add solar body to auction stage and staticinfo, please...

Otherwise just nitpicks

Assets/Scripts/Augment/AugmentImplementations/Hat/HatBarrelModel.cs
Fueredoriku: Yes, i would very much like to so!

I'll do it when i refator attachmentPoints and outputs

toberge (approved):
Sure I suppose it is fine now.

Assets/Scripts/Augment/GunBody.cs
Fueredoriku: There is bad coupling between gunController and augments. I'll refactor that in another PR

Assets/Scripts/Augment/GunController.cs
Fueredoriku: Ye, good point

PR #114, opened 2023-03-15
Fix volume slider scale

toberge:
Make volume sliders adjust volume logarithmically to account for how humans perceive sound volume — logarithmically, not linearly.


Fueredoriku (approved):
Nice fix! :)

Issue #92, opened 2023-02-21
Handle players going out of bounds


toberge:
Killbox or reset y position or sth

Issue #70, opened 2023-02-01
Re-factor AuctionDriver and related architecture


Fueredoriku:
Because its currently ugly

PR #118, opened 2023-03-18
Add basic version of new map

toberge:
- [x] Terrain with mountains and a valley for the arena
- [x] Basic textures
- [x] Basic terrain shader
- [x] Whiteboxed buildings
- [x] Multiple spawnpoints, shuffled

## Progress images:

image

image


Assets/Scripts/Gamestate/MatchController.cs
```diff
@@ -148,7 +148,7 @@ public IEnumerator WaitAndStartNextRound()
         yield return new WaitForSeconds(roundEndDelay);
         // This needs to be called after inputs are set at start the first time this is needed.
         PlayerInputManagerController.Singleton.ChangeInputMaps("FPS");
-        SceneManager.LoadScene("DemoArena");
+        SceneManager.LoadScene("Cratertown");
```
Fueredoriku, line 151: Maybe we should rename this to PascalCase?

Assets/Scripts/UI/StatBar.cs
```diff
@@ -66,6 +66,14 @@ void Awake()
 
     private void UpdateMeter()
     {
+        // Avoid one peculiar bug where the RectTransform is null
+        // and the MissingReferenceException spam prevents the next scene from loading
+        if (baseRect == null)
```
Fueredoriku, line 71: Seems very weird, is there something underlying here or is this just Unity UI jank (tm)?

Fueredoriku (commented):
Very cool! 😎 A nice shuffle algorithm implementation too!

I only have a couple of nitpicks in the review-comments.

Also, I think we should implement a barrier to prevent this, although that can be handled as it's own PR to resolve #92 :
image

Assets/Scripts/Gamestate/MatchController.cs
toberge: Yeah I thought of it but forgor to do it :skull:

Assets/Scripts/UI/StatBar.cs
toberge: I suspect there's an underlying issue, though this is outside the scope of this PR so I just slammed inn a temporary fix :upside_down_face:

Fueredoriku: This isssue is tied to #70 , so we'll keep toss this on there.

Fueredoriku (approved):
Read code and tested, works on my machine 🎉

PR #121, opened 2023-03-19
Make music layers change dynamically

toberge:
- Add general fade and refactor some parts of the music track manager
- Make battle music switch to looped version after intro
- Make string layer in battle music fade in when there are two players left, or there is 30% left on the match timer


Fueredoriku (approved):
Great feature, music transitions sounds great!

My only nitpick is that the log function could automaticaly clamp the 0 values to avoid putting in 0.0001f into each call, but it's not strictly necessary 😅

Code read and tested, I approve!

toberge:
> Great feature, music transitions sounds great!

Thanks! 

> My only nitpick is that the log function could automaticaly clamp the 0 values to avoid putting in 0.0001f into each call, but it's not strictly necessary :sweat_smile:

Yeah I know it's not perfectly clear. Could rewrite that bit now.

PR #123, opened 2023-03-21
Add shacks and rocks to CraterTown

toberge:
image


Fueredoriku (commented):
Nice new models :)

It might be a "me problem" but I get the following error message everytime I enter CraterTown:
image

You should probably check if the same applies to you before merging

toberge:
> It might be a "me problem" but I get the following error message everytime I enter CraterTown: (...)

Yeah I've got the same issue actually, will fix

Fueredoriku (approved):
Great! :)

PR #127, opened 2023-03-21
Improve player movement

toberge:
- Replace opposing force on canceled input with *drag*
  - Different drag in air vs on ground, to prevent jumping from being borked
  - Adjust jump and strafe forces to compensate
- Limit horizontal velocity only, by clamping the x and z velocity of the rigidbody

Please test and provide feedback, I'm not certain that this is how we want it :upside_down_face:


Fueredoriku (approved):
Cleaner code and no weird oppsoite force at stopping? Aww ye 🤠

For my personal taste I like when players walk a bit faster, but that will probably something we can discuss next meeting

Reads good and works on my machine, I approve!

PR #128, opened 2023-03-21
Fix battle theme playing in bidding scene

toberge:
Yeah, I made this kind of mistake *again*.
The coroutine should stop regardless of the kind of track we start playing, otherwise it ends up starting the looping part of the battle theme while we are in the bidding scene. Yikes.


Fueredoriku (approved):
Missed that too in the review 😅

PR #129, opened 2023-03-22
Let players step up stairs

toberge:
currently horrible implementation, test if you want to experience pain


toberge:
Obsolete. We solved this by sloping everything.

PR #130, opened 2023-03-23
Fire extension!

Fueredoriku:
- OnHit implementation not yet added, as we are awaiting some other changes, will be adressed in anoter PR.


toberge (approved):
Looks fire 🔥
We might need to adjust a bunch of things here, and we _do_ need to add actual fire damage, but this seems fine _for now_

PR #134, opened 2023-03-23
Improve bidding UI

toberge:
- [x] Add border in player color around stat UI
- [x] Edit bullet speed label
- [x] Use coin icon for chip count in stats (consistency is key)
- [x] Use same font on bidding platform texts as we do elsewhere

image  
image  
image


Fueredoriku (approved):
Looks great!

Code reviewed and tested :)

PR #135, opened 2023-03-25
Add damage border around player HUD

toberge:
- [x] Tweened red border appears when players are hit
- [x] Border persists at approximately 1/4th health and lower

image


Fueredoriku (commented):
I'm getting some weird errors as single player, haven't yet tested with more:
image

toberge:
> I'm getting some weird errors as single player, haven't yet tested with more: image

Ah, that's because of the way I stuck a custom shader to a canvas; I added a material to an image, noticed those errors and kinda ignored them :))))

Fueredoriku (approved):
Looks very nice, and it's a great addition!

I'm not so sure about those warnings, but since it can be handled in its own separate issue, I'll leave an approve on this

Good work!

toberge:
> Looks very nice, and it's a great addition!
> 
> I'm not so sure about those warnings, but since it can be handled in its own separate issue, I'll leave an approve on this
> 
> Good work!

Added a bogus `_MainTex` property so that it shuts up. Probably required since it's a material on a `RawImage`, but it's not relevant at all for this particular case. Please let me know if there's a better way to add a shader to an UI element than adding an image component and slamming the shader on that one.

Fueredoriku (approved):
Very much approve of this, great work! 😁👍

PR #136, opened 2023-03-27
Explosive barrel added

Davidspilde:
I have added barrel, vfx and code for explosion barrel.
Explosions damage radius should be discussed in the next meeting its not the best in this state


Assets/Scripts/ExplotionController.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using Unity.Mathematics;
+using Unity.VisualScripting;
+using UnityEngine;
+using UnityEngine.VFX;
+
+public class ExplotionController : MonoBehaviour
```
toberge, line 9: Explo**s**ion, not explo**t**ion.

Assets/Scripts/ExplotionController.cs
```diff
@@ -0,0 +1,73 @@
...
+        foreach (Collider collider in colliderList)
+        {
+            HitOrMiss(collider);
+        }
+    }
+
+    private void HitOrMiss(Collider collider)
+    {
+        HitboxController controller = collider.GetComponent<HitboxController>();
+        if (!controller.health || !hitHealthControllers.Contains(controller.health))
```
toberge, line 51: Please check that the controller itself is not null (`!controller`) as well, as I suspect that's what colliders without a HitboxController will return in production mode.

Assets/Scripts/ExplotionController.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            hitHealthControllers.Add(controller.health);
+            float distdam = math.pow((Vector3.Distance(collider.transform.position, transform.position) / radius), 2);
```
toberge, line 54: Please rephrase this variable – I get what you mean, but it could be written in a more clear, verbose way. Abbreviations should be avoided in this context. (and do note that damage is not relevant to this variable, it's more like a scale factor based on distance)

Assets/Scripts/ExplotionController.cs
```diff
@@ -0,0 +1,73 @@
...
+            float damageFinal = (1-distdam)*damage;
+            Debug.Log(damageFinal+": Dette er skaden");
+            Debug.Log("Avstanden er "+ Vector3.Distance(collider.transform.position, transform.position)/radius);
+            Debug.Log("Posisjonene er :"+transform.position+" for Barrel || "+collider.transform.position+" for hitbox");
```
toberge, line 58: Please remove superfluous logs from your PR, we're able to test this through the damage to the players now.

Assets/Scripts/ExplotionController.cs
```diff
@@ -0,0 +1,73 @@
...
+            controller.DamageCollider(damageFinal,gameObject);
+        }
+
+    }
+    
+    
+ 
+}
+
+///TODO:
```
toberge, line 68: TODO comments shouldn't linger in your code once they are completed.

Assets/Scripts/Obstacles/ExplodingBarrel.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        GetComponent<HealthController>().onDeath += Explode;
+    }
+
+    private void Explode(HealthController controller, float damage, DamageInfo info)
+    {
+        barrelMesh.enabled = false;
+        GetComponentInChildren<CapsuleCollider>().enabled = false;
+        GetComponent<ExplotionController>().Explode();
+
```
toberge, line 21: Perhaps this object should destroy itself after its explosion has finished? (this isn't crucial)

toberge (changes requested):
Explosion looks good and damages players. Haven't tested with other barrels nearby, though it seems like that should be handled already. However, you haven't placed the exploding barrel in the arena. Please rebase your branch onto the newest version of `dev` and place a few barrels in the CraterTown scene, so we can see it in action more easily.

I've left a few comments on lines where I've found issues with your code – don't be scared, this is part of the learning process, young padawan. 

As for the balancing issue you mentioned, we should also adjust the barrel's health and damage. The damage seems a little much to me, and the health should be tuned down if we want players to be able to destroy barrels quickly. We'll have to rebalance everything lots of times going forward though, so there's no real problem here.

Assets/Scripts/Obstacles/ExplodingBarrel.cs
Davidspilde: Fixed destroys after 4 seconds so the vfx plays

Assets/Scripts/ExplotionController.cs
toberge: Never mind, turned out to not be necessary since we're limited by layer!

Davidspilde:
Dont merge!
only after sondres damageinfo update

toberge:
> Dont merge! only after sondres damageinfo update

Ready for a merge, this pull request is.

toberge (approved):
> Dont merge! only after sondres damageinfo update

Ready for a merge, this pull request is.

I made some changes here, so somebody else should test this first!

Assets/Scripts/Obstacles/ExplodingBarrel.cs
```diff
@@ -0,0 +1,27 @@
...
+    private void Explode(HealthController controller, float damage, DamageInfo info)
+    {
+        barrelMesh.enabled = false;
+        GetComponentInChildren<CapsuleCollider>().enabled = false;
+        GetComponent<ExplosionController>().Explode(info.sourcePlayer);
+        Destroy(gameObject, 4);
+    }
+}
+
+//Husk hvem som gjorde skade og send det videre
```
Fueredoriku, line 26: Should maybe rewrite to english or delete

toberge: whoops, there was something more I'd missed here :sweat_smile:

PR #146, opened 2023-04-12
Straight firin'

toberge:
Progress:

- [x] Crosshair
- [x] Bullets are fired at a target found by raycast
- [x] Interpolate to original transform at close range
- [ ] Refactor projectile state setting so we avoid doing all the same peculiar stuff in subclasses >_<
  - Doesn't seem like a major concern

Needs testing. Try walking close to another player and shooting them. Does it behave as expected? I suspect not >_>

### DIFF
--- a/Assets/Scripts/Augment/GunController.cs
+++ b/Assets/Scripts/Augment/GunController.cs
@@ -1,9 +1,9 @@
-using System.Collections;
-using System.Collections.Generic;
 using UnityEngine;
 
 public class GunController : MonoBehaviour
 {
+    private const float outputTransitionDistance = 2;
+
     [HideInInspector]
     public ProjectileController projectile;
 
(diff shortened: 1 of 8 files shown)


Assets/Scripts/Augment/GunController.cs
```diff
@@ -60,6 +59,13 @@ private void FireGun()
         }
         stats.Ammo--;
 
+        // Aim at target but lerp in original direction if target is close
```
Fueredoriku, line 62: This should probably be in onFire instead to avoid the raycast getting stuck while holding the fire button

Fueredoriku (changes requested):
Code looks great, and the gameplay is much better! 😎👍

I only request one change in that holding down fire never updates the raycast, creating som weird behaviour like this:
jank

If this is fixed it should be ready to merge!

Assets/Scripts/Augment/GunController.cs
toberge: Good catch, fixed!

Fueredoriku (approved):
Looks great, tasted and works on my machine!

toberge:
> Looks great, tasted and works on my machine!

mmm, delicious pull request

PR #152, opened 2023-04-19
Fix player stat UIs not showing in builds

toberge:
Seems to have to do with setting canvas elements to 0 transparency _or_ disabling them. Fixed with a hacky wait-one-frame solution.


Fueredoriku (approved):
Tested and works on my machine!

Fixed a build error for unset URP asset in all non-ultra quality settings :)

PR #155, opened 2023-04-19
Improved bidding

Fueredoriku:
- [ x ] Remove healthcontroller from bidding players
- [ x ] Display models of augment on bidding platform
- [ x ] Fix bug related bidding in overtime


toberge (approved):
With the replacement of a dangerous delegate = null statement, this pull request is satisfactory in my scrutinising eyes. Good work, Friedrich

PR #156, opened 2023-04-19
Added simple sound system which plays shooting and hitting sounds

TArchetype:


toberge (approved):
Nicely implemented, works great in practice, let's goooooo

Issue #160, opened 2023-04-21
Fire range for bullets seems too low


toberge:
![image](https://user-images.githubusercontent.com/44986729/233633912-43a14953-3c45-4c66-8425-8586bbd32458.png)

would expect to hit from this distance so I leave it as a bug :))))))

PR #161, opened 2023-04-21
Adjust guns

toberge:
- Slug barrel isn't implemented, it's simply a clone of the bullet barrel. **Removed**. 
- Double barrel is broken at the moment, and makes no sense as a barrel with our current design (should be an extension). **Removed**. 
- Bullet range was too low, so I doubled it :smiling_imp: 

Closing #160


Assets/Scripts/Augment/AugmentImplementations/ContinuousDamage.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    private void OnDeath(HealthController healthController, float damage, DamageInfo info)
+    {
+        this.enabled = false;
+    }
+
+    private void Update()
+    {
+        if (hitbox && Time.time - lastDamageTime > 1 / damageRate)
+        {
```
Fueredoriku, line 35: We might want to add a onHit invocation here, but it's not strictly necessary yet

Assets/Scripts/Extensions/TransformExtensions.cs
```diff
@@ -0,0 +1,20 @@
+using UnityEngine;
+
+namespace TransformExtensions
```
Fueredoriku, line 3: Nice use of type extension!

Fueredoriku (approved):
Great fixes/changes!

Added a small commit to fix a bug where players would keep taking continous damage after death.

I would suggest we try to do something about the instatiated fire scale later, but it doesn't need to be in in this PR.

I'll leave an approve here, good work! 😎👍

Assets/Scripts/Extensions/TransformExtensions.cs
toberge: Thanks!

Assets/Scripts/Augment/AugmentImplementations/ContinuousDamage.cs
toberge: That's a little outside the aim of this component, but it seems like the sort of thing we would want for our _actual_ hat barrel implementation

PR #166, opened 2023-04-25
Fix player stat UI not showing stats

toberge:
PlayerStatUIs weren't specifically enabled after a recent fix, so I moved the OnEnable code to an Init method :)


Fueredoriku (approved):
Nice fix 👍

PR #176, opened 2023-04-26
nullrefrenceexception errors

TArchetype:
This request removes the errors we get when one presses subscribed buttons when the options menu is not enabled


toberge (approved):
I approb :+1:

PR #177, opened 2023-04-26
Feature/actual cel shade

SondreHus:
Looks good on my machine


Fueredoriku (approved):
Looks great!

The only nitpicks I have is that mesh particles are missing outlines, but I added a separate issue for this #181 , so we'll fix that later. It could also be nice to have an endlines in the HLSL files, but that is also very nitpicky.

Also: It seems some augment materials are no longer on the objects. I think this is the case for both gatling body and hat barrel. This could be fixed here or in a separate issue.

Tested and read code, I approve!

PR #178, opened 2023-04-26
Add barebones version of electric coil barrel and refactor barrel animation

toberge:


Assets/Scripts/Augment/AugmentImplementations/BarrelAnimator.cs
```diff
@@ -0,0 +1,12 @@
+using UnityEngine;
+
+/// <summary>
+/// Abstract for reload and fire animations.
+/// Not an interface since interface types can't be editor fields :(
```
Fueredoriku, line 5: Unity pain ☢

Assets/Scripts/Augment/BulletController.cs
```diff
@@ -14,34 +16,62 @@ public class BulletController : ProjectileController
...
+
+        // TODO should be in projectilecontroller :)
+        gunController = transform.parent.GetComponent<GunController>();
+        if (!gunController)
+        {
+            Debug.Log("HatBarrel not attached to gun parent!");
+            return;
+        }
+        gunController.onInitializeGun += OnInitialize;
+        //gunController.onFire += OnFire;
```
Fueredoriku, line 47: Why is onFire removed? 🤔

Fueredoriku (approved):
Looks great! 👍 

I only have a small comment, but I've tested and looked through the code, so I'll leave an approve for you 🤠

Assets/Scripts/Augment/BulletController.cs
toberge: It was replaced with a call to InitializeProjectile (and not invoked anymore), at this point we could actually subscribe to that delegate again though >_>
and I may have just copied this over from HatBarrel without putting too much thought into it
and these names give me pain btw :))))

toberge: There, it's cleaned up now... Though please test the game, seems like I broke the hat barrel on my end :/

toberge: ok thanks for fixing my mistake fredrik

PR #187, opened 2023-05-04
Dance Body

Fueredoriku:
Solves #175 

- Added a dance body that reloads when you dance
- Renamed BarrelAnimator to AugmentAnimator

A separate issue to add sounds will be made after merge 🤠

ddrFinal


Fueredoriku:
ddrFinal

Assets/Scripts/Augment/AugmentImplementations/DDRBodyAnimator.cs
```diff
@@ -0,0 +1,24 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class DDRBodyAnimator : AugmentAnimator
+{
+    private Animator animator;
+    public override void OnFire(int remainingAmmo)
+    {
+        animator.SetTrigger("Fire");
```
toberge, line 10: What does it do on fire? What does it even mean in this context? OnFire seems to be called on every beat in DDRBody, which is _more than a little_ confusing. The `BarrelAnimator` class wasn't really meant for bodies in the first place :/

Assets/Scripts/Augment/AugmentImplementations/DDRBody.cs
```diff
@@ -0,0 +1,185 @@
...
+    private const float targetPoint = 3;
+
+    [SerializeField]
+    private Precision[] precisionsGoodToBad;
+
+    LTDescr arrowMover;
+
+    [SerializeField]
+    private float secondPerArrow = 1;
+    private float musicPaces = 0.4285714f;
```
toberge, line 46: Singular pace?

Assets/Scripts/Augment/AugmentImplementations/DDRBody.cs
```diff
@@ -0,0 +1,185 @@
...
+        precisionText.color = precision.Value.color;
+        precisionText.gameObject.LeanScale(new Vector3(1.1f, 1.1f, 1.1f), 0.5f)
+            .setEasePunch()
+            .setOnComplete(
+            () => precisionText.enabled = false);
+            
+        LeanTween.value(gameObject, SetFlashFactor, 0, 20f, 0.5f).setEasePunch();
+
+        gunController.Reload(reloadEfficiencyPercentage*precision.Value.awardFactor);
+        animator.OnReload(1);
```
toberge, line 108: So um... Why do you call `OnReload` with a bogus argument? Perhaps this only happens because there_is_ an argument in the `OnReload` method in `BarrelAnimator`. Does it make that much sense to use an integer for this on a gun body, when it doesn't even care about the magazine size?

toberge (changes requested):
Haven't playtested and didn't intend to go over everything right now, but I've left a few comments.

Assets/Scripts/Augment/AugmentImplementations/DDRBody.cs
Fueredoriku: The problem with this is two-fold:
- There might (likely) exist barrels thad doesn't care about magazineSize, but wants to do something OnReload
- I believe BarrelAnimator can and should be generalized to an augmentAnimator with optional parameters instead

I didn't want to make such a change without talking it through, and especially not in this PR due to it's already large size

Assets/Scripts/Augment/AugmentImplementations/DDRBodyAnimator.cs
Fueredoriku: While I agree the trigger name itself ("Fire") might be a bit misleading (It's animating the membrane to vibrate when doing many gun functions, and to the beat), I think we should have a general augmentAnimator class

Assets/Scripts/Augment/AugmentImplementations/DDRBody.cs
```diff
@@ -0,0 +1,185 @@
...
+    [SerializeField]
+    private AugmentAnimator animator;
+
+    private Material ddrMaterial;
+
+    private const int screenMaterialIndex = 2;
+
+    private float arrowHeigth = 0;
+
+    private int arrowDirectionQuadrant = 0;
```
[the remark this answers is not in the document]
Fueredoriku: I agree!

Assets/Scripts/Augment/AugmentImplementations/DDRBody.cs
Fueredoriku: Will fix!

Assets/Scripts/Augment/AugmentImplementations/DDRBodyAnimator.cs
Fueredoriku: I updated the name to something less vague than "Fire", which I think will suffice for now

toberge (approved):
With the changes you made, it should be fine to merge this boi

PR #188, opened 2023-05-06
Feature/non lego guns

SondreHus:
Haha, you have to spend your fucking saturday reviewing a pull request. Enjoy my new and "improved" vfx system boyos.
Peace!


Assets/PinAnimator.cs
```diff
@@ -0,0 +1,33 @@
...
+    [SerializeField]
+    private float delay = 0f;
+
+    [SerializeField]
+    private AnimationCurve easeCurve;
+    void Start()
+    {
+        var projectileController = GetComponentInParent<GunController>();
+
+        projectileController.onFire += (GunStats stats) => recoil();
```
toberge, line 22: Please let Recoil() take GunStats as argument so we can += and -= the method instead

toberge (approved):
The new barrels look great, the shotgun is an actual shotgun, and all seems fine and dandy except for the console exceptions and the one thing I commented on. I'm fine with merging this in and fixing the exceptions later, especially since one of them is the same ol' IndexOutOfRangeException in the VFX formatter (the texture might be too small or sth, I haven't checked) and thus not created by this PR.

This set of exceptions is new though:
image
(in BulletController::Start())

Assets/PinAnimator.cs
toberge: Resolved this myself. Made this class actually use the BarrelAnimator that I already wrote.

Fueredoriku (approved):
Code looks good, seems to work when tested

I do however get a lot of errors:
image
They might be a me-problem, and I think we can toss it into the tecnical debt pile, but we shoudl look into this later.

Good work though, I approve! 🤠👍

PR #191, opened 2023-05-06
Feature/post match scoring

### LINKED ISSUE #171: Reward screen after completed shooting round

ThomasJGabrielsen:
Pull request contains a post match scoring system.

Closes #171 

scoreboard


Fueredoriku:
Bidding is broken due to playerIdentity somehow no longer being transported from FPS to bidding. This must be fixed before approve

image

toberge (changes requested):
A few major usability issues:

- The death screen for the last death in a round doesn't show. We should make it clear who died and allow the winner to style on them.
- In a similar vein, there is no delay before switching to bidding. We should wait a little before going there.

Assets/Scripts/Gamestate/MatchController.cs
```diff
@@ -140,15 +179,6 @@ public void EndActiveRound()
         roundTimer.OnTimerUpdate -= HUDTimerUpdate;
         roundTimer.OnTimerRunCompleted -= EndActiveRound;
         AssignRewards();
-
```
Fueredoriku, line 143: This should not have been moved from matchManager into TownScoreBoard. The reason it was structured this way initialy was to ensure the all functions subscribed to the onRoundEnd delegate were invoked before we changed the scene. 

Now important functions like the one that decouples and transfers playerInput isn't able to be executed before players scene is changed. This leads to the bidding scene being broken.

Assets/Scripts/Gamestate/MatchController.cs
```diff
@@ -140,15 +179,6 @@ public void EndActiveRound()
...
-        if (!IsWin())
```
Fueredoriku, line 144: Removing this makes it impossible to win a match.

Fueredoriku (changes requested):
This is a nice feature, and the panning is really smooth! 👍

There are some issues that need to be adressed, mainly that the way you call/use MatchController breaks parts of the game, so I would personally try to implemement the call to TownScoreBoard on the end of a round as an async operation that is await-ed in MatchController instead.  

Also the "crime" naming scheme for variables in TownScoreBoard and Poster classes is a bit confusing to me.
The scoreboard in the scene should also be made into a prefab.

This branch also need to be rebased before merge.

ThomasJGabrielsen:
Closed for major overhaul

PR #195, opened 2023-05-10
Added cratertown volume, fixe cfx indexing and coil gun

SondreHus:


toberge (approved):
Yes looks very good I believe in you

PR #202, opened 2023-05-11
Add lightning trail to coil gun

toberge:
With spiral extension:
coil spiral lightning

With new muzzle flash:
coil flash


Assets/Scripts/Augment/BulletModifiers/ZigzagPathModifier.cs
```diff
@@ -0,0 +1,60 @@
...
+        while (Vector2.Dot(random, lastOffset) > 0)
+        {
+            random = Vector2.Perpendicular(random);
+        }
+        return random;
+    }
+
+    public void AddZigzagDisplacement(float distance, ref ProjectileState state)
+    {
+        if (state.distanceTraveled > interval + (float)state.additionalProperties[lastDistanceId])
```
Fueredoriku, line 32: Why not use state.distanceTraveled directly? 
The we wouldn't need the lastDistanceId additionalProperty

Fueredoriku (commented):
Looks great! ⚡
~~I notice HDR has been bumped a bit, also for non-coil stuff like explosions and solar body:
I highly approve of this change, but I suggest we mention it has been done somewhere in the PR~~

I also have a comment regarding the lasteDistanceId additionProperty, if we resolve that thread I will put up an approve! 🤠👍

toberge:
> Looks great! ⚡ I notice HDR has been bumped a bit, also for non-coil stuff like explosions and solar body: I highly approve of this change, but I suggest we mention it has been done somewhere in the PR

Sondre added HDR volume in a separate pull request

Assets/Scripts/Augment/BulletModifiers/ZigzagPathModifier.cs
toberge: The problem is that we can't tell when we passed the interval with just the distance travelled, since projectiles jump in unpredictable chunks from the modifier's point of view. Could try calculating last step from other variables and use that to determine when we enter a new interval, but that'll make it a more convoluted check

Fueredoriku: > The problem is that we can't tell when we passed the interval with just the distance travelled
But isn't that what you do in the code though? 
Specifically: `state.additionalProperties[lastDistanceId] = state.distanceTraveled;` in the body of the if statement

It seems like the only place it's updated? 🤔

toberge: It's the distance travelled last time the lightning turned, which is only updated when the if statement triggers.

Fueredoriku: I see now that the value is a timestamp and not a refference, so the implementation is very sensible 👍

Fueredoriku (approved):
Code looks good, runs like expected, and looks amazing!

Here's an approve from me! 👌

PR #208, opened 2023-06-14
Second battle theme: Machiavellian Mayhem

toberge:
- [x] New new battle theme
- [x] Randomly pick battle theme
- [x] BPM and beats per bar info in music tracks

PR #209, opened 2023-06-14
Sync DDR body with the music

toberge:
~~requires #208 to be merged first~~ (done!)

# what I wrote about this problem

> ok så når du får til ei pil
> så starter neste pil i y=0
> rett etter at du traff den forrige pila
> så hele greia havner ut av synk
> og det er ikke bare varigheten opp og helt ut av skjermen som må matche musikken, det er også når den treffer... hitsonen eller hva en nå kaller det

# solution

The arrow moves at a fixed pace based on the beats per bar and beats per minute of the currently playing music track. It is supposed to hit the mark at the start of each bar (this is up for debate, try it and see if you can get a feel for the "rhythm").
When the arrow is reset, it behaves as shown in the diagram below:

```
                    if arrow ends up here:
screenHeight ______ -> reset to 0
             |    |
targetHeight | <> | -> reset to startHeight = -(screenHeight - targetHeight)
             |    |
             |    | -> reset to lower than when it hits the mark
             |    |
             |    |
             |____| where arrow starts:
           0        -> starts here when out of bounds

startHeight         -> starts here when it hits the mark (and initially)

< startHeight       -> starts here when it is triggered too early
```

### DIFF
--- a/Assets/Scripts/Augment/AugmentImplementations/DDRBody.cs
+++ b/Assets/Scripts/Augment/AugmentImplementations/DDRBody.cs
@@ -1,5 +1,3 @@
-using System.Collections;
-using System.Collections.Generic;
 using TMPro;
 using UnityEngine;
 using UnityEngine.InputSystem;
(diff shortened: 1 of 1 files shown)


Fueredoriku (approved):
Seems to work well! 🤠👌

I think this is definitely more polished, and the code looks good, so I'll leave an approve 👍

PR #210, opened 2023-06-29
Improved reload for solar body

Fueredoriku:
The solar body now actually considers where the sun is, and if there are any obscuring objects to block the sun, making for more intuitive reload mechanics

solarBetterReload

### DIFF
--- a/Assets/Scripts/Augment/AugmentImplementations/Solar/SolarBody.cs
+++ b/Assets/Scripts/Augment/AugmentImplementations/Solar/SolarBody.cs
@@ -61,9 +61,9 @@ public class SolarBody : GunBody
 
     void FixedUpdate()
     {
-        Vector3 relativeRotationSolarpanel = rayCastOrigin.transform.TransformDirection(Vector3.up);
-        float cosineSimularity = Vector3.Dot(relativeRotationSolarpanel, globalLightDirection.eulerAngles);
-        if ((!Physics.Raycast(rayCastOrigin.position, globalLightDirection.eulerAngles, maxObscuringCheckDistance, obscuringLayers.value)) && cosineSimularity > 0)
+        Vector3 solarPanelPlaneRotation = rayCastOrigin.transform.TransformDirection(Vector3.up);
+        float orientationOverlap = Vector3.Dot(solarPanelPlaneRotation, globalLightDirection.eulerAngles);
+        if ((!Physics.Raycast(rayCastOrigin.position, globalLightDirection.eulerAngles, maxObscuringCheckDistance, obscuringLayers.value)) && orientationOverlap > 0)
         {
             Reload(gunController.stats);
         }
--- a/Assets/Scripts/Augment/AugmentImplementations/Solar/SolarBody.cs
+++ b/Assets/Scripts/Augment/AugmentImplementations/Solar/SolarBody.cs
@@ -23,6 +23,7 @@ public class SolarBody : GunBody
 
     private const float maxObscuringCheckDistance = 15f;
     private const float coolDownSeconds = 0.5f;
+    private const float chargeUpSeconds = 1; 
     private bool isCooldown = false;
 
     private const int solarPanelMaterialIndex = 3;
@@ -47,6 +48,10 @@ public class SolarBody : GunBody
     {
         if (isCooldown)
             return;
+        if (solarPanelMaterial.GetFloat("_On") == 0)
+        {
+            LeanTween.value(gameObject, SetEmissionStrength, 0, 1, chargeUpSeconds);
+        }
         solarPanelMaterial.SetFloat("_On", 1);
         gunController.Reload(reloadEfficiencyPercentagen);
         isCooldown = true;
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Augment/AugmentImplementations/Solar/SolarBody.cs
```diff
@@ -24,6 +31,8 @@ public override void Start()
     {
         meshRenderer.materials[solarPanelMaterialIndex] = Instantiate(meshRenderer.materials[solarPanelMaterialIndex]);
         solarPanelMaterial = meshRenderer.materials[solarPanelMaterialIndex];
+        // The main directional light in the scene (the sun) should ideally be tagged as such, but this works for now
+        globalLightDirection = FindFirstObjectByType<Light>().transform.rotation;
```
toberge, line 35: Could you specify `DirectionalLight` instead of just `Light`?

Assets/Scripts/Augment/AugmentImplementations/Solar/SolarBody.cs
```diff
@@ -50,9 +59,11 @@ private IEnumerator CoolDown()
         isCooldown = false;
     }
 
-    void Update()
+    void FixedUpdate()
     {
-        if (transform.parent.rotation.eulerAngles.y < sunDirection + sunDirectionSpan && transform.parent.rotation.eulerAngles.y > sunDirection - sunDirectionSpan)
+        Vector3 relativeRotationSolarpanel = rayCastOrigin.transform.TransformDirection(Vector3.up);
```
toberge, line 64: SolarPanel, and the whole thing should be relativeRotationOfSolarPanel if you're this verbose.
Couldn't you just use the forward vector of the origin if it is a point you define yourself, or is it a reference to the actual solar panel plane?

Assets/Scripts/Augment/AugmentImplementations/Solar/SolarBody.cs
```diff
@@ -50,9 +59,11 @@ private IEnumerator CoolDown()
...
+        float cosineSimularity = Vector3.Dot(relativeRotationSolarpanel, globalLightDirection.eulerAngles);
```
toberge, line 65: *Sim**i**larity*, and perhaps call it `overlap` or something more descriptive.

toberge (changes requested):
Looks good! I just have a few comments regarding the code, and I haven't tested it yet but the changes don't look too significant.

Assets/Scripts/Augment/AugmentImplementations/Solar/SolarBody.cs
Fueredoriku: The directional light class is unintuively not the Light we're looking for (Its part of an experimental global lighting package), and the light we want is a Light object instance with a directional property.

Assets/Scripts/Augment/AugmentImplementations/Solar/SolarBody.cs
toberge: https://docs.unity3d.com/ScriptReference/Transform-up.html :)

toberge:
Names look good but I'd still suggest writing `transform.up` and thest this before I approve

toberge (changes requested):
So in addition to that one comment I have reopened: After briefly testing the solar body, I do wish we could lerp between the two states because the transition from animated grid to pure darkness feels quite jarring. Other than that it works great, now it's actually usable, while not entirely OP since other players can bait the solar wielder into dark areas.

Fueredoriku:
I agree that lerp-ing to reload state is nice, but after testing I've concluded that lerping to the inactive state from active removes visual clarity in terms of showing new users what conditions turns the reload off. I kept the sudden change for gameplay concerns in that specific case and added lerp for the other :)

Assets/Scripts/Augment/AugmentImplementations/Solar/SolarBody.cs
```diff
@@ -8,14 +8,22 @@ public class SolarBody : GunBody
...
     [SerializeField, Range(0, 1)]
     protected float reloadEfficiencyPercentagen = 0.1f;
 
-    private const float sunDirection = 90f;
-    private const float sunDirectionSpan = 45f;
+    private Quaternion globalLightDirection;
+
+    private const float maxObscuringCheckDistance = 15f;
     private const float coolDownSeconds = 0.5f;
+    private const float chargeUpSeconds = 1;
```
toberge, line 26: When the lerp goes on for a whole second I can see why it would not make the point of change clear to the user, but you could lerp for like 100ms or less and it should still be noticeable :)

Though I'm not going to prevent you from pressing the big green button this time.

toberge (approved):
it's probably fine now, you fixed things

Assets/Scripts/Augment/AugmentImplementations/Solar/SolarBody.cs
Fueredoriku: I tested with separate values for both, but yeah imma merge lol

PR #212, opened 2023-07-21
Ammo hud and updated healthbar

Fueredoriku:


toberge (changes requested):
Seems to respond correctly to ammo usage, however this is how it looked when I had one shot left in the hatbow:
image
And when depleting chunks of the bar, it sometimes lines up with the gaps between the holes and sometimes not.

Fueredoriku:
>it sometimes lines up with the gaps between the holes and sometimes not.
That is intentional due to varying amounts of ammo used per shot

What I can du though, if need be, is to remove 90 degrees (the last quadrant behind the healthbar) to always show the bar unless empty

Fueredoriku:
The ammoBar when reloading and using the starter body looks a bit weird because the auto-reloads as soon as its magazine is depleted. I want to remove that behaviour in a future PR, so please disregard that specific gun combo looking a bit weird at the moment 😅

Assets/Scripts/UI/PlayerHUDController.cs
```diff
@@ -36,6 +36,7 @@ public class PlayerHUDController : MonoBehaviour
     private Material damageBorder;
 
     private const float ammoSpinDegrees = 30;
+    private const float availableCircumference = 270;
```
toberge, line 39: Why not available _degrees_?

toberge, line 39: Why not available _degrees_?

toberge (approved):
You fixed the main issue and now the starter body has reload so... Go ahead and merge it (perhaps after considering my renaming suggestion)

Assets/Scripts/UI/PlayerHUDController.cs
toberge: Why is this duplicated smh

Fueredoriku: Yea, I agree that might be a better name, git push -f incoming

PR #213, opened 2023-07-31
Fix mesh projectiles getting caught on owners pan extension

Fueredoriku:
Adjusted pan hitbox and output as to not block user's own mesh projectiles


toberge (approved):
Unconventional branch name but it seems to work. Had one instance of a hat sticking to the barrel (???) but that's probably unrelated.

PR #214, opened 2023-08-01
Revolver body reload

Fueredoriku:
Adding a slooooow reload animation for the starting body so it isn't better than all the other bodies in the game 

basicReload

I plan to further improve this animation in the future by making the player "cram" bullets into the opened slot during the animation, but this is something I want to implement together with actual "gun holding" IK animations. (Will be a whole other PR)


toberge (approved):
It's great to finally have some actual reload for the basic body, even though it's a little placeholder-y.

The animation looks a little _empty_ when you're just staring at the barrel for some seconds, but meh, that'll be fixed later. The smoke is a nice touch.
Looks a little off with the WIP cog barrel of course:
image
Other than that, it is now very noticeable that the shotgun has too much ammo, we should consider modifying ammo cap in barrels :)))))

PR #215, opened 2023-08-04
Improved movement

Fueredoriku:
Based on feedback from playtesting with colleagues

- Added falling acceleration
- Removed velocity clamp on grounded state
- Increased movement speed

### DIFF
--- a/Assets/Scripts/Control&Input/PlayerMovement.cs
+++ b/Assets/Scripts/Control&Input/PlayerMovement.cs
@@ -27,7 +27,7 @@ public class PlayerMovement : MonoBehaviour
     private float maxVelocity = 10;
 
     [SerializeField]
-    private float strafeForce = 50;
+    private float strafeForce = 60;
 
     [SerializeField]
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Control&Input/PlayerMovement.cs
```diff
@@ -44,6 +44,8 @@ public class PlayerMovement : MonoBehaviour
     [SerializeField]
     private float jumpTimeout = 0.5f;
 
+    private const float marsGravity = 3.72076f;
```
toberge, line 47: Not sure if realism in gravity constant is what we want to prioritize

toberge (approved):
Feels snappier, surely is an improvement

Assets/Scripts/Control&Input/PlayerMovement.cs
Fueredoriku: Agreed, we might tweak this value later when we move to more vertical level design I think, but this feels OK for now I think

PR #217, opened 2023-08-16
Weapon holding

### LINKED ISSUE #158: Fix aiming edgecase
image

idk what is going on here, might just require adjustment of lerp distance

Fueredoriku:
- A general 2 body inverse kinematics implementation
- Inverse kinematics for weapon holding
- Fixes #158 

genericIKFinalFinal


Assets/Scripts/Gamestate/PlayerManager.cs
```diff
@@ -149,7 +156,8 @@ private void UpdateAimTarget(GunStats stats)
     {
         Vector3 cameraCenter = inputManager.transform.position;
         Vector3 cameraDirection = inputManager.transform.forward;
-        if (Physics.Raycast(cameraCenter, cameraDirection, out RaycastHit hit, maxHitDistance, hitMask))
+        Vector3 startPoint = cameraCenter + cameraDirection * targetStartOffset;
```
toberge, line 159: Does this create issues when players are facing off real close?

Assets/Scripts/Utils/GenericTwoBodyIK.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// </summary>
+    /// <param name="root">First bone</param>
+    /// <param name="joint">Second bone, will extend towards pole when squished</param>
+    /// <param name="end">Third and final bone, will always point to target</param>
+    /// <param name="pole">The half-sphere direction which the second bone points from</param>
+    /// <param name="target">The goal point of the end bone</param>
+    /// <param name="rootOffset">Optional extra rotation of root around its IK forward axis</param>
+    /// <param name="jointOffset">optinal extra rotation of joint around its IK forward axis</param>
+    /// <param name="localScaleOffset">Optional relative scaling for achieving distances comparable </param>
+    public static void AnimateTransforms(Transform root, Transform joint, Transform end, Transform pole, Transform target, float rootOffset = 0, float jointOffset = 0, float localScaleOffset = 1)
```
toberge, line 22: So you ended up writing your own IK code, then? Unity didn't have anything that fit?

toberge (approved):
Much better than floating guns, but the elbow joints don't seem to move in a sensible way, the arm flips when you aim downwards, and the gun clips through your shoulder:
image

The reload animation makes it look a little odd as well:
image

And solar + crosshat... what is he holding?
image

Same problem with dancer + shotgun:
image

Assets/Scripts/Gamestate/PlayerManager.cs
Fueredoriku: Yes, but only when the gun is so close that its "inside" the other player, at which point I think it's fine anyway 😅

Assets/Scripts/Utils/GenericTwoBodyIK.cs
Fueredoriku: Nope, they have a 2bodyIK component, but that only works with humanoid rigs (Our player uses a generic rig), and for some reason Unity has no "generic" alternative 🙃.

Fueredoriku:
I tweaked the hand bone to rotate better with the gunbodies. 😁👍
The rest of the visual quirks either require more computations than they are worth / some would require manually implementing constraints. 

Most of this can be mitigated in the current system should future problems arise by further tuning provided offset values, existing animations, or even updating some of the gunbodies to have better scaling or dropping two-handed carrying if we decide we don't want arm clipping through body at all.   

A fix altogether could be to have different models in first person compared to what others see, which would make for less extreme gun position distances relative to the player. This feature would also inherently solve a lot of the other aforementioned quirks, but we'll have to evaluate if such a feature should be prioritized later 😅

For now I'll merge this in!

PR #218, opened 2023-08-17
Make DDR loudspeaker unz to the beat

toberge:
- [x] Make Unz lead to new Unz in the animation controller
- [ ] Adjust Unz animation speed to the music
- [ ] Offset Unz animation if necessary
- [ ] Prevent Unz on fire and reload, or
- [ ] Do a big Unz on fire, while maintaining the pace

image


toberge:
Won't rework this. May end up doing this later but it requires some funky animation timing :sweat_smile:

PR #220, opened 2023-08-27
Add bidding music

toberge:
- Add remixed version of menu theme as bidding theme
- Tweak menu music so that it is louder and has greater audio quality

Closing #219


Fueredoriku (approved):
Sounds great!! 😁

Not only is this very nice for the more frantic bidding, but we can reuse the old theme in the new planned cutscene where the auctioneer presents items before the bidding starts as a cool transition, so it's great that you left it in!

PR #223, opened 2023-08-30
Animated chip for arenas

Fueredoriku:
A spinning chip prefab to be used in future level design, spesifically for new members so it can be placed in levels to plan level design.

To help with onboarding of new members, I've left out the implementation of all non-aesthetic functionality.

collectableCoin


toberge (approved):
chip it :shipit:

PR #243, opened 2023-10-15
Feature/platform movement

karofmah:


toberge (changes requested):
The platforms seem to work really well and don't glitch out when stomping the player :+1: 

However, I do have some minor and major comments about the code quality.

Assets/Scripts/Gamestate/MatchController.cs
```diff
@@ -94,7 +94,7 @@ void Start()
         // Makes shooting end quickly if testing with 1 player
 #if UNITY_EDITOR
         if (PlayerInputManagerController.Singleton.playerInputs.Count == 1)
-            roundLength = 5f;
+            roundLength = 60f;
```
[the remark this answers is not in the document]
Fueredoriku: Great idea, I'll put up an issue 👍

Assets/Scripts/Platforms/PlatformCollision.cs
```diff
@@ -0,0 +1,13 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+public class PlatformCollision: MonoBehaviour
```
toberge, line 4: Hold on, I don't see this component on any of the platforms in crater town, is it not in use?

toberge: Oh nevermind it's on the one that moves up and down – so it isn't required on the others?

karofmah: It is attached to a child of the platform named "Platform". The purpose of the child is to vertically reverse the movement of the platform when colliding with a player that is below the platform.

Fueredoriku (approved):
Great work! 😁👍

Assets/Scripts/Platforms/PlatformCollision.cs
toberge: And thus it isn't necessary for platforms that move horizontally? I'd imagine there could be cases where platforms should reverse horizontally as well, so renaming it to something _vertical_ doesn't seem necessary. However, perhaps it could be renamed to reflect its bounce-back intent? (the platforms already have collisions)

Assets/Scripts/Platforms/PlatformMovement.cs
```diff
@@ -0,0 +1,66 @@
...
+     if (currentDistance < 2 || currentDistance + 2 >= travelDistance)
+    {
+        moveSpeed = 2f;
+
+    }else{
+        moveSpeed = 5f;
+    }
+
+    if (nextRoutepointIndex != routepoints.Count) return;
+        routepoints.Reverse();
```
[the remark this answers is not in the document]
toberge: The indent is still here :upside_down_face:

Assets/Scripts/Platforms/PlatformCollision.cs
karofmah: I can rename it, but do you think that it is realistic that a player is stomped by a platform that moves horizontally? All platforms does reverse when reaching the end, but the purpose of the script is to reverse stomping players/avoid unintended interactions.

toberge (approved):

Assets/Scripts/Platforms/PlatformCollision.cs
toberge: Nonono that's not what I meant :sweat_smile: It's just that I only saw this component on the vertically moving platform and started thinking about whether it was general or not. I _do_ think that players should not be stomped by platforms, and that if there's a horizontal platform that closes in on a player that is falling down in the gap between the platform and its destination, we should avoid potential shenanigans by reversing it. The name sounds better now :+1:

PR #246, opened 2023-10-15
Add background to main menu

ChrisKrane:


toberge:
image
The gallery doesn't work 🤔

Assets/Scripts/UI/MainMenu/MainMenuTumbleweed.cs
```diff
@@ -0,0 +1,34 @@
...
+        // Rotates model 360 degrees 
+        model.transform.LeanRotateAroundLocal(model.transform.forward,360,1).setLoopCount(-1);   
+
+
+    }
+
+    private void LateUpdate()
+    {
+        // Get the height of the terrain at the position of TumbleweedParent
+        terrainHeight = terrain.SampleHeight(transform.position);
```
Fueredoriku, line 28: I have suspicion that running this function every update is pretty costly (depending on if Unity samples their terrain from a heigthmap of some sort or by actually calculating geometry). 
Might not be a problem as there isn't much other perfomance heavy stuff at all in the main menu, but we might have to monitor this a bit

Fueredoriku (commented):
Great work, looks nice 😁👍

I think we should take a look at fixing the gallery before merging (maybe just do it as easy resetting the camera position to its old position when switching to gallery (and then back to you new position when changing back from the gallery) if thats what's causing this? Fixing the underlying problem for the gallery is not part of the scope of this PR I would say, so I would approve a quickfix like that)

Assets/Scripts/UI/MainMenu/MainMenuTumbleweed.cs
toberge: Since the terrain is created by painting on a heightmap, this should devolve down to sampling that texture and thus perform well.

Assets/Scripts/UI/MainMenu/MainMenuMoveCamera.cs
```diff
@@ -0,0 +1,36 @@
...
+    private float cameraSpeed;
+    [SerializeField]
+    private float rotationAngle;
+    [SerializeField]
+    private Transform cameraPositionDefault;
+    [SerializeField]
+    private Transform cameraPositionAlternative;
+
+    // Start is called before the first frame update
+    void Start()
```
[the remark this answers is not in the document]
ChrisKrane: Fixed

ChrisKrane:
> Great work, looks nice 😁👍
> 
> I think we should take a look at fixing the gallery before merging (maybe just do it as easy resetting the camera position to its old position when switching to gallery (and then back to you new position when changing back from the gallery) if thats what's causing this? Fixing the underlying problem for the gallery is not part of the scope of this PR I would say, so I would approve a quickfix like that)

Fixed with a small tweak to GalleryMenu.cs :)

Fueredoriku (approved):
Very nice, great first PR 🤠👍

Issue #238, opened 2023-10-05
DDR body material should be self-lit


Fueredoriku:
Screen currently becomes invisible in abscene of light

Issue #228, opened 2023-09-23
DDR arrows sometimes doesn't appear


Fueredoriku:
The arrow sometimes stop appearing after som point in a round. This resets when a new round starts.

~Issue might have something to do with controller drifting?~

Issue can be replicated by spinning the stick in circles for a while

PR #247, opened 2023-10-16
Fix DDR issues

Fueredoriku:
- Fixed DDR arrow sometimes disappearing due to not being reset on gamepads under certain conditions #228 
- Fixed DDR screen material not emitting any light, causing it to be completly dark with lack of lighting #238 
Here is the material still being visible in the *tumbleweed chamber*
image

### DIFF
--- a/Assets/Shaders/ddrShader.shadergraph
+++ b/Assets/Shaders/ddrShader.shadergraph
@@ -4171,9 +4171,9 @@
     "m_ShaderOutputName": "Emission",
     "m_StageCapability": 2,
     "m_Value": {
-        "x": 0.0,
-        "y": 0.0,
-        "z": 0.0
+        "x": 0.037735819816589358,
+        "y": 0.037522222846746448,
+        "z": 0.037522222846746448
     },
     "m_DefaultValue": {
         "x": 0.0,


Assets/Scripts/Augment/AugmentImplementations/DDRBody.cs
```diff
@@ -126,6 +127,9 @@ private void Fire(InputAction.CallbackContext ctx)
 
     private void ArrowSelect(InputAction.CallbackContext ctx)
     {
+        if (arrowHeight < lowestCheckHeight)
+            return;
```
toberge, line 131: Nice catch! I started wondering why the arrows were missed before they even show up.

This makes me wonder why we don't have any visual response when we register a _miss_, though (as opposed to the arrow being ignored and going all the way up)

toberge (approved):
Nice fixes! Haven't tested it yet but doesn't look like anything should break :)))))

Assets/Scripts/Augment/AugmentImplementations/DDRBody.cs
Fueredoriku: The main reason is that players would have stand completly still in order to not always get misses when playing with a gamepad

toberge: Aha, understandable

toberge: I do think we could have some sort of detection of when players actively flick the stick, but that sounds too fancy for just a miss indicator

PR #256, opened 2023-10-21
Improve slope movement

### LINKED ISSUE #255: Slope movement
Players should move _along_ slopes :)

toberge:
- Use same boxcast as ground check
- Find slope normal
- Use slope normal if angle is low enough

Closes #255


Fueredoriku (commented):
Nice fix! 👍 

I suggest we set the slopeAngleTreshold to 50 instead maybe? I think 60 is a bitt too much (see gif)

Also this reminds me that we could slap on some IK for the player legs, but that'll have to be a discussion we take another time (and in another issue)

climber

Fueredoriku (approved):
Very nice 🎉

PR #260, opened 2023-10-21
Normalize auctioneer voicelines

### LINKED ISSUE #258: Auctioneer voice drowns in the music
We should up the volume of the auctioneer's voice lines (at least normalize them).

toberge:
This makes his magnificent voice audible, though not necessarily legible.

Closes #258


toberge:
Ideally (some of) these should be replaced but at least they're _audible_

Fueredoriku (approved):
Better to test on a demo with audible voices, indeed

PR #262, opened 2023-10-21
Fix exception on solo suicide

### LINKED ISSUE #261: Suicide leads to exception when soloing

toberge:
Replace First with FirstOrDefault to avoid exception when there's no one else alive.

Closes #261


CraZyB1336 (approved):
Looks good

PR #263, opened 2023-10-22
Feature/grand canyon modeling

CraZyB1336:
Finished most modeling of the grand canyon/grand crevice map (name yet to be decided)

What needs to be done:
- Add more props 
- Model houses 
- Add interact-able props (coins, barrels, etc)


toberge (approved):
This map looks amazing now!

Only major issue is perhaps that the death zone is quite low down so players take a while to fall :/

PR #272, opened 2023-10-24
Fix spawnpoint placement

Fueredoriku:
SpawnPoint 3 and 4 in grand crevice has been added to the PlayerFactory.
All spawnpoints in bidding scene moved into camera FOV


toberge (approved):
Tested, works well.

I do need to mention that the spawn points in Crater Town (not covered in this PR) could use some rotation to prevent players from seeing a wall when they spawn in 😅 Perhaps they all could look towards the church?

PR #275, opened 2023-11-05
Fix solar reload and firerate

Fueredoriku:
- Directional lights in all scenes now tagged with "mainLight"
- Lowered solar firerate
- Created more accurate check for sun coverage for solar
- Fixed barrel hitbox in cratertown


CraZyB1336 (approved):
Looks nice! Good to finally have a working solar body hehe.

toberge (approved):
Seems fine, it's still a machinegun but no longer OP :)))))

PR #276, opened 2023-11-05
Fix crouching being more visible, fix gamepad look sensitivity

Fueredoriku:
- Fix crouching being more visible
- Fix gamepad look sensitivity
- Made players accelerate to the top walking speed quicker

### DIFF
--- a/Assets/Scripts/Control&Input/PlayerMovement.cs
+++ b/Assets/Scripts/Control&Input/PlayerMovement.cs
@@ -130,10 +130,11 @@ public class PlayerMovement : MonoBehaviour
 
         if (ctx.performed)
         {
+            if (IsInAir())
+                return;
             animator.SetBool("Crouching", true);
             strafeForce = strafeForceCrouched;
-            if (!IsInAir())
-                inputManager.gameObject.LeanMoveLocalY(localCameraHeight - crouchedHeightOffset, 0.2f);
+            inputManager.gameObject.LeanMoveLocalY(localCameraHeight - crouchedHeightOffset, 0.2f);
         }
             
         if (ctx.canceled)
--- a/Assets/Scripts/Control&Input/PlayerMovement.cs
+++ b/Assets/Scripts/Control&Input/PlayerMovement.cs
@@ -132,7 +132,7 @@ public class PlayerMovement : MonoBehaviour
         {
             animator.SetBool("Crouching", true);
             strafeForce = strafeForceCrouched;
-            if (!IsInAir())
+            //if (!IsInAir())
                 inputManager.gameObject.LeanMoveLocalY(localCameraHeight - crouchedHeightOffset, 0.2f);
         }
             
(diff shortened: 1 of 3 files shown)


toberge (changes requested):
image
The weapon is lowered before the character, is this intentional?
And it feels like the gamepad sensitivity is a little low now – though this should be adjustable in the end :)))))

Assets/Scripts/Control&Input/PlayerMovement.cs
```diff
@@ -198,6 +198,7 @@ private void UpdatePosition(Vector3 input)
                     var groundNormal = GroundNormal();
                     var direction = Vector3.ProjectOnPlane(input, groundNormal);
                     body.AddForce(direction * strafeForce, ForceMode.VelocityChange);
+                    body.AddForce(direction * strafeForce, ForceMode.Impulse);
```
toberge, line 201: also excuse me what? why is there an extra impulse force to the regular walking?

Assets/Scripts/Control&Input/PlayerMovement.cs
```diff
@@ -132,15 +132,15 @@ private void SetCrouch(InputAction.CallbackContext ctx)
...
+            //if (!IsInAir())
                 inputManager.gameObject.LeanMoveLocalY(localCameraHeight - crouchedHeightOffset, 0.2f);
         }
             
         if (ctx.canceled)
         {
             animator.SetBool("Crouching", false);
             strafeForce = strafeForceGrounded;
-            if (!IsInAir())
+            //if (!IsInAir())
```
toberge, line 143: What's going on with these commented-out checks that make execution always fall through to moving the camera?

Assets/Scripts/Control&Input/PlayerMovement.cs
Fueredoriku: Adressing feedback regarding acceleration when starting to move being too low. This change helps movement feel more "snappier".

I forgot to declare it in the PR description, will do it now

Assets/Scripts/Control&Input/PlayerMovement.cs
Fueredoriku: Forgot to remove properly, but the short story is that I found an undetected bug where players would be "locked" into crouching-heigth if crouch jumping and unpressing crouch while in air

The only reson we didn't see this bug earlier was that the crouching heigth was small enough for it not to be apparent

Fueredoriku: Is now updated 👍

toberge (changes requested):
Tested again and
- The crouching animations are still not in sync 🙃 
- Crouching midair prevents the camera from being lowered, even after you land on the ground:

Fueredoriku:
> * The crouching animations are still not in sync 🙃
It cant be and has in fact never been in sync. As long as the player camera and gun is coupled like they are atm, I deem this as a separate issue (more related to layer separation)
> * Crouching midair prevents the camera from being lowered, even after you land on the ground:
This is intentional, as Crouching and uncrouching in air looks very bad, only uncrouching does not, which is why I left it in this way. I did however notice that the isair check shoukd be called earlier so crouchspeed isn't "locked" when jumping, I pushed up a commit for this now.

toberge (approved):
Removing the ability to crouch while in the air seems like a good change.
The desync of the camera and crouching animations do become more prevalent with this PR, though I suspect this may be worsened by how fast the player enters the crouching state in the animation.

PR #278, opened 2023-11-07
Fix hat barrel

Fueredoriku:
- Fix hat materials
- Add outlines to mesh particles!!
- Made hats ricochet and bounce around for a while
   - Still gets stuck in either player, or in the terrain given it has traveled long enough
BouncyHats2

### DIFF
--- a/Assets/Scripts/Augment/AugmentImplementations/Hat/HatBarrel.cs
+++ b/Assets/Scripts/Augment/AugmentImplementations/Hat/HatBarrel.cs
@@ -34,6 +34,8 @@ public class HatBarrel : ProjectileController
     [SerializeField]
     private VisualEffect hatVfx;
 
+    [SerializeField]
+    private float maxLengthBeforeStuck = 100;
     protected override void Awake()
     {
         base.Awake();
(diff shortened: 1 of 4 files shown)


Assets/Scripts/Augment/AugmentImplementations/Hat/HatBarrel.cs
```diff
@@ -34,6 +34,8 @@ public class HatBarrel : ProjectileController
     [SerializeField]
     private VisualEffect hatVfx;
 
+    [SerializeField]
+    private float maxLengthBeforeStuck = 100;
```
toberge, line 38: Distance instead of length? And add a newline before Awake?

Assets/Scripts/Augment/AugmentImplementations/Hat/HatBarrel.cs
```diff
@@ -136,13 +138,23 @@ private void UpdateProjectile(ProjectileState state)
 
         if (collisions.Length > 0)
         {
-            state.active = false;
-            OnColliderHit?.Invoke(collisions[0], ref state);
             HitboxController hitbox = collisions[0].GetComponent<HitboxController>();
```
toberge, line 141: TryGetComponent instead?

toberge (approved):
Seems to work quite well and I especially like the bouncing motion 🏈

PS: The hats still get really small from the viewpoint of the player that fired them and this bugs me, but that's not created by this PR anyways

Assets/Scripts/Augment/AugmentImplementations/Hat/HatBarrel.cs
Fueredoriku: Ye, I agree

Assets/Scripts/Augment/AugmentImplementations/Hat/HatBarrel.cs
Fueredoriku: That would be better to do while I'm at it, I agree

PR #279, opened 2023-11-07
Add functionality to collectable chip, add chips to levels

Fueredoriku:
New and improved chip!
- Better textures (now featuring bump maps, selective metallicness and prettier albedo texture)
- Collectable chips placed around in both maps
- Collectable chips can now in fact be collected
newChip


toberge (approved):
Looks alright, and it is great to have some other way to collect chips, but as Thomas stated, the chips are a little plain. I'll approve this so I won't have to press approve after some small adjustments 🙃  I also find the lack of any effect on pickup a little offputting, and we really need a chip counter, but that's not _too_ important right now.

toberge:
Seems like there's a brief moment where the player is actually stopped by the chip collider before it disappears 🤔

Fueredoriku:
> I like the new chip design. However i feel it does not look too nice on maps when the directional light is directly above, see below. bilde

This is a problem with the directional light in that specific stage being set by multiple light sources (which is a bit jank)

> Perhaps adding a wobble on the axis on the coin as it revolves would help. In addition adding a random initial rotation would prevent the coins from sharing identical rotations. Perhaps adding some sort of particle effect would also help the coins "pop" more out from their surrondings.

I don't agree with this due to wanting to emulate the movement of coins in 3D mario games.

I will override this request as these changes should be added asap in its current state due to: time constraints before spillexpo and potential scene merge conflicts.

toberge (approved):
The bumpin' is gone 👍

PR #281, opened 2023-11-09
Auction polish

Fueredoriku:


toberge (changes requested):
Looks really good, I just have two suggestions (as mentioned on Discord):
- We should have a (radial) progress bar for the time, as people have been confused about it
- The description could be squished vertically to leave more space for other elements as the descriptions tend to be very short
  - This would leave more space for the timer

toberge (approved):
The radial timer works really well, nice addition!

I do notice one thing that's been around for a while: There's a substantial delay at the end of the bidding session where absolutely nothing happens, could we perhaps lower biddingEndDelay in the MatchManager prefab? To like just a second or two? 👉 👈

PR #285, opened 2023-11-11
Add body that reloads with ammo pickups

toberge:
Ammo pickups spawn when someone has this body:
Ammo

Reload by collecting them (they aren't collected if your magazine is full):
AmmoPickup

Players without the ammo pickup body can't see the ammo boxes:
image

Closing #286


Fueredoriku:
> Players without the ammo pickup body can't see the ammo boxes

Players with this gunpart suffers from schizophrenia 😔✊

Assets/ScriptableObjects/Items/GatlingBody.asset
```diff
@@ -14,6 +14,6 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   displayName: Gatling
   secretName: humongous
-  displayDescription: A fast and heavy hitting body
+  displayDescription: Huge magazine but you need ammo boxes to reload
```
Fueredoriku, line 17: Maybe rewrite to "Huge magazine, but needs ammo boxes to reload"

Assets/Scripts/Augment/AugmentImplementations/AmmoBoxCollector.cs
```diff
@@ -0,0 +1,35 @@
+using System.Collections;
+using UnityEngine;
+
+public class AmmoBoxCollector : MonoBehaviour
+{
+    private const string ammoBoxBodyName = "Gatling";
```
Fueredoriku, line 6: Should be exhanged with a serializeField refference to the corresponding Item scriptableobject and just compare  it like so: `hasAmmoBoxBody = player.identity.Body == ammoBoxBodyItem;`

Fueredoriku (approved):
Looks great, I think this will be a great addition! 🤠👍

I have just a couple of requests:
- I think there are too many ammoboxes in grand crevice (this augment may be a bit too strong atm), maybe tune it down a little bit either in amount of ammoboxes or Ammo size or both 
- The check for if a player hast the gatling item should be done with a refference to the "gatling" Item scriptableObject instead of gatling by name

Everything else seems good to me, so I'll leave you an approve to use at your own discretion ;)

PR #287, opened 2023-11-12
Move death box up in Grand Canyon

toberge:
This lessens the time spent falling before players actually die there.

⚠️ This might mess with players trying to fly underneath the map using the knockback extension


CraZyB1336 (approved):
Wow, ikke 75 straight? 😬 

Neida, dette ser og føles bedre ut.

PR #290, opened 2023-11-13
Chip feedback and small Auction polish

Fueredoriku:
- Add audio for collecting chips
- Add audio for auctioneer sales
- Fix lighting in grand canyon for better chip look there
   - Changes skybox and upps light intensity for better bounced lighting, removed extre directional light from scene (no longer needed)
- Add curly curtain material to auction
image

TODO
- Someone, at some point, must bake the new lightmap in the grand crevice scene


Assets/Scripts/Auction/Auctioneer.cs
```diff
@@ -12,6 +12,8 @@ public class Auctioneer : MonoBehaviour
     [SerializeField]
     private AudioSource audioSource;
     [SerializeField]
+    private AudioSource audioSourceFX;
```
toberge, line 15: This is for the kaching sound? Please reconsider the name

Assets/Scripts/Interactables/CollectableChip.cs
```diff
@@ -22,6 +24,8 @@ private void OnTriggerEnter(Collider other)
         if (!other.gameObject.TryGetComponent<PlayerManager>(out PlayerManager player))
             return;
         player.identity.UpdateChip(1);
-        Destroy(gameObject);
+        audioSource.Play();
+        GetComponent<Collider>().enabled = false;
+        chipModel.SetActive(false);
```
toberge, line 29: You could still destroy the coin after the audio has played

Assets/Scripts/Auction/Auctioneer.cs
```diff
@@ -65,6 +67,7 @@ public void Sell()
         audioSource.Stop();
         audioSource.clip = sold.RandomElement();
         audioSource.Play();
+        audioSourceFX.PlayDelayed(0.5f);
```
toberge, line 70: This delay seems just a little too long, perhaps shorten it by 100 ms?

Assets/Scripts/Gamestate/PlayerManager.cs
```diff
@@ -266,7 +266,7 @@ private void SetLayerOnSubtree(GameObject node, int layer)
 
     private void PlayOnHit()
     {
-        if (Random.Range(0, 100) > 5)
+        if (Random.Range(0, 1000) > 5)
```
toberge, line 269: bye bye colonel kernel

toberge (changes requested):
The curtains look nice and it's good to have audio on chip pickup and purchase, but the volume is so low it drowns in the music so could we turn it up a little?

Assets/Scripts/Gamestate/PlayerManager.cs
Fueredoriku: 😔✊

Assets/Scripts/Interactables/CollectableChip.cs
Fueredoriku: Ye, but I think the cost of running an IEnemurator waiting for the sound to finish playing is basicly the same as disabling the object like this, so I think this is the better solution for the brevity.

Assets/Scripts/Auction/Auctioneer.cs
Fueredoriku: Will do 👍

Assets/Scripts/Auction/Auctioneer.cs
Fueredoriku: Sure :)

Assets/Scripts/Interactables/CollectableChip.cs
toberge: Destroy(gameoObject, audioSource.clip.length)
(usikker på om det er clip)
Resolve hvis du ikke vil prøve

Assets/Scripts/Auction/Auctioneer.cs
toberge: one should be voicelineSound and the other buySound or sth more actually meaningful but I'm not gonna fite you on dis

toberge: oh actually do that because you funked up and didn't set the audiosource after rename >:)

toberge (approved):
leaving an approve so you can merge after fixing the stuff. sound sounds audible now

karofmah (approved):
Looks great! Really like the audio for collecting chips

PR #292, opened 2023-11-14
Add procedural animations for augments

Fueredoriku:
- Add jiggling for ddr and solar
- Add a weapon punch when firing the gun


toberge (approved):
Looks great! The fire punch adds some much-needed feedback to the parts that don't have an existing fire animation 🥊 and the wire jiggle makes them look less like steel wire :)))))

PR #293, opened 2023-11-14
Fix/start button not working

ChrisKrane:
Fixes missing variable names in mapNames array in MainMenuController that lead to start button not working.


toberge:
The button works, nice 👍

toberge (approved):
The button works, nice 👍

PR #294, opened 2023-11-14
Weapon construction

Fueredoriku:


Assets/Scripts/Auction/AuctionDriver.cs
```diff
@@ -38,6 +38,28 @@ public class AuctionDriver : MonoBehaviour
 
     [SerializeField]
     private Animator cameraAnimator;
+    public GameObject Camera => cameraAnimator.gameObject;
+
+    public static AuctionDriver Singleton;
+
+private void Awake() {
```
toberge, line 45: This method should be indented one level.

Assets/Scripts/Control&Input/ItemSelectManager.cs
```diff
@@ -0,0 +1,300 @@
...
+        timer.OnTimerRunCompleted += ChangeScene;
+        timer.OnTimerRunCompleted += SetLoadout;
+        timer.OnTimerRunCompleted -= UpdateTimer;
+
+
+    }
+
+    private void ChangeItemDisplayed(GameObject previousItem, GameObject nextItem,Transform itemSpawnPoint){
+
+        if(previousItem != null && nextItem != null){
```
toberge, line 109: This `if` should be inverted so you return early instead

Assets/Scripts/Control&Input/ItemSelectManager.cs
```diff
@@ -0,0 +1,300 @@
...
+    private Image bodySelect;
+    [SerializeField] 
+    private Image barrelSelect;
+    [SerializeField] 
+    private Image extensionSelect; 
+    [SerializeField]
+    private Color selectedColor;
+    [SerializeField]
+    private Color defaultColor;
+    private int selectedIndex = 0;
```
toberge, line 57: So this is for switching between types of parts, and should not have a generic name since we have indices for individual parts too. Perhaps call it `selectedTypeIndex` or `partTypeIndex`?

Assets/Scripts/Control&Input/ItemSelectManager.cs
toberge, line 1: This file should be reformatted and simplified a little, won't comment too much on specifics now as it'll probably have to go in like this for Spillexpo :))))))))

toberge (approved):
Finally we get to build our weapons 🎉 
The stats seem to display correctly too 📊 

image

Now, I do see a bunch of things to complain about in the current iteration of this UI, but most (if not all) can be left until after Spillexpo:

- The timer is displayed on every screen involved in the process. I think we could do it similarly to the timer in the FPS rounds.
- The gun's name is displayed, but no description of what the augments actually _do_, which might be an issue for some players, and this kinda makes the stats more important than the fancy functionality we add to the guns
- The up/down arrows don't fit the style of RPR, this isn't a pixel art game last I checked 😉 
- Some gunparts are severely obscured by the UI background, like the pan extension:
image

toberge:
for the record: just resolve my comments and merge this, the comments aren't that important

Assets/Scripts/Auction/AuctionDriver.cs
karofmah: Done!

Assets/Scripts/Control&Input/ItemSelectManager.cs
karofmah: Fixed it!

Assets/Scripts/Control&Input/ItemSelectManager.cs
karofmah: I changed it to selectedTypeIndex.

Assets/Scripts/Control&Input/ItemSelectManager.cs
karofmah: I agree this file should be refactored. I will do it after Spillexpo :)

PR #296, opened 2023-11-15
Feature/post match scoring

ThomasJGabrielsen:
Adds a scoreboard at the end of the round to tally the points players get.
bilde

Known issues/missing improvements:
- Does not cover coins picked up in a round, these are automatically added to savings.
- Text is slightly difficult to read for certain player colors.
- There is no skip functionality implemented to skip the tallying process.
- Scoreboard has no animations, may seem a bit to static
- Image is redundant


toberge (commented):
Nice work! 🎉 Now we'll finally be able to see our cowbots' life savings and blood-stained chips 🤠 

However, I do have some concerns:
- There is no delay after the victor has been determined, possibly making it hard to notice who won in a close-cut standoff as you no longer see the "killed by player x" screen for the last two standing. Consider adding a delay of a couple of seconds before transitioning to the scoreboard animation.
- The chip entries are displayed very inconsistently:
  image
  Ideally they should stick to the bottom and have consistent spacing.

Overall it should be ready for Spillexpo, but we could squeeze in some polish.

toberge:
Revamped look:
image

toberge (approved):
I would've liked to review this properly but it seems fine now (and I adjusted the scoreboard UI and delays a little). The choice of "crime" as the term for scoreboard reward entries seems odd and I'd like to see this refactored someday, but we ain't got time for that now 😔

PR #298, opened 2023-11-15
Feature/sodiepopper

SondreHus:
Added the poppin' sloppin' body

Did the essential tweaking, but the reaload delay and firerate might be too "good". To make realoading "harder" increase the threshold, but be wary, as the system is not currently consistent mathematically, but good enough as long as the threshold is not too large.


toberge (approved):
Works great, very nice!

Assets/Animation/SodiePopper.meta
toberge, line 1: Meta: Don't merge before fredrik has fixed the remaining buggeroonies

PR #305, opened 2023-11-15
Feature/remodel cubes in crater town

CraZyB1336:
Added Models and Material textures to Crater Town.
image


Fueredoriku (approved):
Looks great 😁🎉

Amazing improvements to the map

PR #309, opened 2023-11-16
Changed reference to deprecated class

ThomasJGabrielsen:


CraZyB1336 (approved):
Wow

toberge (approved):
Side note: It isn't deprecated, just renamed

PR #310, opened 2023-11-17
Add model to seeker augment and misc fixes

Fueredoriku:
- Add seeker model
- Cleaned up statUI a bit
- Made chips unlit until we fix the lightning in grand canyon


CraZyB1336 (approved):
Looks Good!

CraZyB1336 (changes requested):
Actually shall we change player placement on CharacterBackground Prefab first?
image

With these position stats it seems (atleast player1) is center

toberge (approved):
Works great! Haven't tested with multiple gamepads but I assume this was tested thoroughly on spillexpo so 👍 

I see you also included the suggestion for the character background positioning, which looks... alright. The 2nd player is off-center, but I'd rather see an overhaul of this screen than have us spend lots more time adjusting its current T-posing placeholder.
image

PR #317, opened 2023-11-25
Make displayed stats match actual stats and adjust ranges

### LINKED ISSUE #183: Shooting VFX is sometimes get out of range error
It seems that setValue call from bullet controller needs to be clamped 🤔

image

toberge:
- Fix displayed shotgun projectile count
- Projectile speed for all barrels, particularily the cannon which has reduced speed (dunno what to do with the hitscan barrels here)
- Refactored collision sample calculation to let us specify samples per unit distance instead of total sample amount, so that one only needs to increase the max distance in order to actually increase the range.
- Increased the ranges of all hitscan weapons, including the shotgun.
  - 100 units is just enough to hit things on the other side of both maps, as long as they aren't mere dots in the distance. Should be sufficient.

May actually resolve #183 but I'm not entirely sure

### DIFF
--- a/Assets/Prefabs/GunParts/BulletBarrel.prefab
+++ b/Assets/Prefabs/GunParts/BulletBarrel.prefab
@@ -630,9 +630,8 @@ MonoBehaviour:
   projectileRotation: {x: 0, y: 0, z: 0, w: 1}
   stats: {fileID: 0}
   player: {fileID: 0}
-  maxDistance: 40
-  collisionSamples: 80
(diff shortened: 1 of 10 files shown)


Fueredoriku (approved):
Changes looks good, and I like the projectilespershot refactor!

I feel the new shotgun range is a tiny bit too long, but I also think we can playtest it, just something to keep in mind.

I'll leave an approve for you to use at your own discretion 👍

PR #320, opened 2023-11-25
Players aren't rotated like the spawnpoints???

toberge:
Tried adjusting the spawnpoints' orientations, but it just doesn't affect the player at all and I am confused


toberge:
Fixed along with #357

Issue #126, opened 2023-03-21
Refactor Augments


Fueredoriku:
- Augments should handle attachmentpoints and outputs instead of gunfactory
- Body should additionaly hold a reload function and a fire-rate setter
- augment should probably have projectile specific functions (it already kinda does, but I believe this will give cleaner code)
- Augment implementations like hatbarrel should probably extend the general barrel class

PR #348, opened 2023-11-27
Revolver reload upgrade

Fueredoriku:
Add player hand that crams bullets into revolver at reload

misc:
- Fix seeker naming for gatling body


Assets/Scripts/Gamestate/PlayerManager.cs
```diff
@@ -227,10 +230,10 @@ public void SetLayer(int playerIndex)
         // Set layers for the camera to ignore (the other players' gun layers, and this layer)
         // Bitwise negation of this player's model layer and all gun layers that do not belong to this player
         // Gun layers are 4 above their respective player layers.
-        var playerAndGunMask = ((1 << playerLayer) | ((1 << (playerLayer + 4)) ^ allGunsMask));
+        var playerAndGunMask = ((1 << playerLayer) | allGunsMask) & ~(1 << (playerLayer + 4));
 
-        // Ignore ammo boxes if this player doesn't have the gatling body
-        var hasAmmoBoxBody = identity.Body.displayName == "Gatling";
+        // Ignore ammo boxes if this player doesn't have the required body
+        var hasAmmoBoxBody = identity.Body == ammoMaskItem;
```
toberge, line 236: thanks, I forgor 💀

Assets/Scripts/Gamestate/PlayerManager.cs
```diff
@@ -227,10 +230,10 @@ public void SetLayer(int playerIndex)
         // Set layers for the camera to ignore (the other players' gun layers, and this layer)
         // Bitwise negation of this player's model layer and all gun layers that do not belong to this player
         // Gun layers are 4 above their respective player layers.
-        var playerAndGunMask = ((1 << playerLayer) | ((1 << (playerLayer + 4)) ^ allGunsMask));
+        var playerAndGunMask = ((1 << playerLayer) | allGunsMask) & ~(1 << (playerLayer + 4));
```
toberge, line 233: Aha, guess I didn't test this enough to see the buggerino -_-
... though I'm not sure what (if anything) was wrong here, care to elaborate? The gun mask should become [bits in gun mask] except [player's gun bit] with the player gun XOR all guns expression 🤔

Assets/Scripts/Augment/PlayerHand.cs
```diff
@@ -0,0 +1,20 @@
...
+using UnityEngine;
+
+public class PlayerHand : MonoBehaviour
+{
+    [SerializeField]
+    private SkinnedMeshRenderer handMaterial;
+    [SerializeField]
+    private GameObject mesh;
+
+    public void SetPlayer(PlayerManager player)
```
toberge, line 12: I feel like this "setter that MUST be called elsewhere after a certain point" pattern is something we should consider avoiding... Perhaps we could create a delegate to serve the same purpose?

Fueredoriku: The purpose is to avoid issues with script execution order, but yes a delegate would definetly serve the same purpose (in a better way too). I would say this issue is very tied to #126, so I'll leave this as is for now I think. (Makes it easier for me when searching for all SetPlayer functions when refactoring later 😅)

Assets/Scripts/Gamestate/PlayerManager.cs
Fueredoriku: It is the same, I changed it while testing and forgot about it 😅

Since you wrote the original, I'll let you decide which version we should keep (whatever we feel is most readable I guess)

toberge: I do of course feel like my own version makes the most intuitive sense, but regardless of which on we go for I do think we should extract the bitshifting expressions to their own variables so that the whole expression becomes easier to read.

Assets/Scripts/Augment/PlayerHand.cs
toberge: Yeah no need to fix it in this PR, I just came up with an improvement for this very thing that has annoyed me for a while

Assets/Scripts/Gamestate/PlayerManager.cs
```diff
@@ -230,7 +230,9 @@ public void SetLayer(int playerIndex)
         // Set layers for the camera to ignore (the other players' gun layers, and this layer)
         // Bitwise negation of this player's model layer and all gun layers that do not belong to this player
         // Gun layers are 4 above their respective player layers.
-        var playerAndGunMask = ((1 << playerLayer) | allGunsMask) & ~(1 << (playerLayer + 4));
+        var playerMask = 1 << playerLayer;
+        var gunMask = (1 << (playerLayer + 4)) ^ allGunsMask;
```
toberge, line 234: Oh, I meant `1 << (playerLayer + 4)` specifically, but this works too. Perhaps we could put of the OR-ing to the `var negatedMask =` line?

Assets/Scripts/Gamestate/PlayerManager.cs
```diff
@@ -33,6 +33,9 @@ public class PlayerManager : MonoBehaviour
     public delegate void BiddingPlatformEvent(BiddingPlatform platform);
     public BiddingPlatformEvent onSelectedBiddingPlatformChange;
 
+    [SerializeField]
+    public Item ammoMaskItem;
```
toberge, line 37: private?

toberge (approved):
omg so cute
image

ok so the reload animation is REALLY GOOD & I love the way the bullets are carelessly dropped and then hammered in
however
tt does look a little weird with the cannon barrel:
image
but we can ignore that :)))))

Assets/Scripts/Gamestate/PlayerManager.cs
toberge: Responded in another comment

Assets/Scripts/Gamestate/PlayerManager.cs
Fueredoriku: Yeah, programming at 3 am moment, will fix 😅

Assets/Scripts/Gamestate/PlayerManager.cs
Fueredoriku: Will do that too 👍

PR #349, opened 2023-11-27
Add bullet hole decals

### LINKED ISSUE #343: Decal bullet hole system
Weapons should ideally leave marks on the world, like bullet holes or explosion marks. 

This should probably be implemented with the decal system.

toberge:
- [x] Add modifier that leaves a decal on hit
- Add decals for
  - [ ] Bullets (bullet, shotgun and coil barrel
  - [ ] Hats
  - [ ] Explosions (cannon barrel)
- [ ] Find and fix edgecases

Closes #343


Assets/Scripts/Augment/BulletModifiers/DecalModifier.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        // Place the decal some distance away from the target so that it is more likely to be projected onto a surface
+        var depthOffset = decal.size.z * depthOffsetFraction;
+        var position = target.ClosestPoint(state.position) - state.direction * depthOffset;
+
+        var spawnedDecal = Instantiate(decal, position, state.rotation);
+        spawnedDecal.transform.parent = target.transform;
+
+        // TODO optionally force it to point down towards the target (requires surface normal)
+        // (this should fix the look of explosions at least)
```
Fueredoriku, line 58: Maybe we should clamp the angle of which these decals can be projected at the surface from no matter the projectile type?
We might be able to avoid creating too long decal lines (like picture below) when the state.direction converges towards being parallel to the surface it is projected on by clamping. This does of course need the surface normal, but the ricochet modifier has some normal aproximation code I wrote that might just do the trick in this case.

image

Fueredoriku (changes requested):
Really nice work, I think the decals really add to the game 😁👍

I have a couple of suggestions:

I think we can safely scale up the explosion decal 3 or even 4 times the size it currently is given the size of explosions, see pic below
image

I would also suggest using the normal approximation code I wrote for the hat ricochet:
`
Physics.Raycast(state.oldPosition, state.direction, out RaycastHit hitInfo);
var surfaceNormal = hitInfo.normal;
`
I think using that would be "good enough" for most decal normal needs 🤠

toberge:
Hold on I actually did so normal adjustment, but I forgor to push it 💀

Assets/Scripts/Augment/BulletModifiers/DecalModifier.cs
toberge: Wellllll ummmm check the updated code 😅

Fueredoriku (changes requested):
Needs to be fixed:
It seems the wrong chip was given the ignore layer, it needs to be set on the collecatableChip and not the UI/chip.

Misc (doesn't need to be fixed):
I also feel like the cannon explosion decal specifically could be even larger still, but I won't die on that hill 😅

Lastly I found something weird (which doesn't necessarily need to be fixed): the explosion decal from exploding the barrel that my crosshair hovers over projects a decal on the ground far below as you can see in the screenshot
image

PR #354, opened 2023-11-29
Add soundFX for revolver reload

Fueredoriku:


Assets/Scripts/Interactables/ExplodingBarrel.cs
```diff
@@ -16,6 +18,7 @@ private void Explode(HealthController controller, float damage, DamageInfo info)
         barrelMesh.enabled = false;
         GetComponentInChildren<CapsuleCollider>().enabled = false;
         GetComponent<ExplosionController>().Explode(info.sourcePlayer);
+        audioSource.Play();
```
toberge, line 21: Why not put this in ExplosionController so we won't have to add the same bit of code to the cannon barrel's explosion?

PR #360, opened 2023-12-03
Use more prefabs in menu scene and fix music

toberge:
- Fix missing music track by *using the prefab of the music track manager that was already made*
- Extract menu pages as prefabs for easier merge conflict handling in the future


Fueredoriku (approved):
Strangely it still seems to be there, but now only when 1 player has joined. Given that we decided to upgrade this part of the menu before next release anyway, I think we can ignore it for now.

I'll leave an approve for you to use at your discretion 🤠👍

toberge:
> Strangely it still seems to be there, but now only when 1 player has joined. Given that we decided to upgrade this part of the menu before next release anyway, I think we can ignore it for now.
> 
> I'll leave an approve for you to use at your discretion 🤠👍

Huh, you sure it isn't also a problem in dev?
and yeah I should've just done the music separately 😔

PR #364, opened 2023-12-04
Add ceiling to arenas

toberge:
And bounds to Grand Canyon


Fueredoriku (approved):
Great change! 
I really like that players on the ground now can reach the ceiling with their shots too (preventing stalling a bit at least 😅)

toberge:
> Great change! I really like that players on the ground now can reach the ceiling with their shots too (preventing stalling a bit at least 😅)

Ideally we'd do something different for the bullets so they phase through the invisible walls, but that's a problem for another time

PR #365, opened 2023-12-04
Reparent camera to head on death

toberge:
For extra ragdoll shenanigans

(don't try this in VR)

deathcam2
deathcam


Fueredoriku (approved):
Code looks good and the feature is very funny! 👌
I think it would be even better in third person, but we can set up a new issue and add that feature another time, I'll leave an approve for now👍

toberge:
> Code looks good and the feature is very funny! 👌 I think it would be even better in third person, but we can set up a new issue and add that feature another time, I'll leave an approve for now👍

Yeah, added #366 for that

Issue #338, opened 2023-11-26
Seeker augment radar


Fueredoriku:
The seeker augment should have a radar that always points towards the nearest available ammobox

Some misc fixes that are also needed:
- Proper antenna jiggle
- Rename from "gatling" to "seeker"

PR #367, opened 2023-12-07
Polish augments

Fueredoriku:
- Add scrolling hacking material on fire for hacking extension screen #321 
- Add and change some wanted scoreboard phrases
- Add antenna wobble to seeker augment
- Add radar to seeker augment #338 
- Add more lerp to rubber sniper, fix momentum bending being in the correct coordinate system
- Add new and better jiggle to augment wires
sekkerRadar


toberge (changes requested):
Two things:
- The radar turns very abruptly when it has a new target, this should ideally be lerped
- The hacking extension is kinda tiny when it's all the way on the edge of the gun... but that's not something we can fix when we have such lomg parts. It does make me wish we had some alternate placements for extensions.
image

Assets/Scripts/Augment/AugmentImplementations/AmmoBox.cs
```diff
@@ -24,6 +26,23 @@ private void Start()
...
+    {
+        ammoBoxes.Add(this);
+    }
+
+    private void OnDisable()
+    {
+        ammoBoxes.Remove(this);
+    }
+
+    public static AmmoBox GetClosestAmmoBox(Transform from)
```
toberge, line 39: This method could take a position vector, it doesn't use anything but `from.position`

Fueredoriku: Done 👌

toberge (approved):
Nice upgrades!

PR #369, opened 2023-12-09
Replace exploding barrel model and adjust stats

### LINKED ISSUE #339: New explosive barrel
The barrel needs to look more "explody"

The barrel health should also be lowered

Optional:
Add explosion sound

toberge:
- Replace model with a new, oil barrel-esque one that looks more inherently dangerous
- Adjust stats so you don't fly quite as high on knockback
- Use cylinder mesh for collider
- Add rigidbody so players can jump off the barrels (!)
- Add the old barrels as regular props
- Reposition the exploding barrels (some were left levitating after crater town fixups)

image

Closes #339


Fueredoriku (approved):
Looks and plays great! 👍

Issue #162, opened 2023-04-22
Bullet VFX indexing issue


toberge:
Sometimes the following appears in the console:
```
IndexOutOfRangeException: Index was outside the bounds of the array.
VFXTextureFormatter.setValue (System.Int32 index, UnityEngine.Vector3 value) (at Assets/Scripts/VFXPositionFormatter.cs:35)
BulletController.InitializeProjectile (GunStats stats) (at Assets/Scripts/Augment/BulletController.cs:94)
GunController.FireGun () (at Assets/Scripts/Augment/GunController.cs:71)
GunController.FixedUpdate () (at Assets/Scripts/Augment/GunController.cs:38)
```
and then no bullet trail is shown :/

PR #370, opened 2023-12-09
Upgrade coil barrel

### LINKED ISSUE #283: Improve coil barrel model
Needs more details:
- [ ] Techy details
- [ ] Charge bar that displays ammo
- [ ] Improved animation?

toberge:
- [x] Make a fancier model
- [ ] Texture it?
- [x] Make fancier animation
- [x] Rewire gun events so burst fire is possible
- [x] Fix bugs caused by rewiring
  - [x] Fix compilation errors from merge with culled gun
  - [x] Fix bullet barrel not firing???
  - [x] Move when targeting happens (target is set in PlayerManager and used in GunController)
  - [x] Prevent third-person gun firing independently from first-person gun (???)
- [ ] Add battery indicator (optional)
- [ ] Add effect from being shot by the barrel that makes player spazz out

WIP blenderin':
image

Closes #283 and #162 (more than before)


Fueredoriku (changes requested):
I have identified two problems that needs to be resolved before merge:

- Muzzleflashes no longer viewable by other players 
- It seems the displayGun hat ammo is no longer being culled (see image below)
image

toberge:
Oh noes :(
Will squash the buggerinoes

toberge:
Fixed & ready for rereview

Fueredoriku (changes requested):
Some new bugs that must be adressed
- Lawnmower is broken (can't fire due to lacking animator)
- Might be a local thing from my end, but cannonBarrel projectiles doesn't render

Also:
Sodie popper + Coil = infinite ammo, but sodie popper needs some rework (as it already has problems with other existing augments) so that can a be it's own PR.

Assets/Scripts/Augment/ProjectileController.cs
```diff
@@ -136,6 +136,11 @@ protected virtual void Awake()
             Debug.Log("Barrel not attached to gun parent!");
             return;
         }
+        if (!gunController.player)
+        {
+            Debug.Log("Barrel w/o player reference, is this a display gun?");
```
Fueredoriku, line 141: I think we should remove this debug so ut doesn't clutter the consoler everytime we enter the FPS scenes

Fueredoriku (commented):
I also notice displayGun no longer recoils, it was a bit subtle before, but I don't think we should remove it

Assets/Scripts/Augment/GunController.cs
```diff
@@ -33,18 +33,42 @@ public class GunController : MonoBehaviour
...
+
     private void FixedUpdate()
     {
         if (fireRateController == null)
         {
             // No fireratecontroller exists
             return;
         }
-        if (fireRateController.shouldFire(triggerPressed, triggerHeld))
+        if (!isFiring && fireRateController.shouldFire(triggerPressed, triggerHeld))
```
Fueredoriku, line 71: I managed to enter a state in which I couldn't fire the game anymore using DDR body + shotgun (seemingly because I reloaded at a bad timing, it's a bit hard to replicate)

toberge:
> I also notice displayGun no longer recoils, it was a bit subtle before, but I don't think we should remove it

I do think we should remove it from most guns but I'll remove it from this PR at least :))))))
Perhaps we could shake the screen a lil instead?

Assets/Scripts/Augment/ProjectileController.cs
toberge: Done.

toberge:
> Some new bugs that must be adressed

o no

>     * Lawnmower is broken (can't fire due to lacking animator)

Fixed. Lots of shenanigans related to ammo deduction and fire rate.

>     * Might be a local thing from my end, but cannonBarrel projectiles doesn't render

That's all on your end. Open the VFX graph again and save it, after that it should work.

> Also: Sodie popper + Coil = infinite ammo, but sodie popper needs some rework (as it already has problems with other existing augments) so that can a be it's own PR.

Yeah we need to deal with the recoil tween later hahahaahha

Assets/Scripts/Augment/GunController.cs
toberge: Oh dear god why :)))))

toberge: Unable to reproduce, could you test again?

toberge:
recoil animation makes DDR body disappear from the screen with coil barrel... gotta fix somehow

Assets/Scripts/Augment/GunController.cs
Fueredoriku: I just managed to reproduce it:
Sometimes we get an VFXformatter out of range exception in BulletController (Which is an old bug).
Before this change, when this exceptions occurs, it seems the gun shot trails aren't rendered properly in that shot.
With this change, since the compiler exits the function before it's finished when the exception occurs, the isFiring is never changed back, permanently disabling the shotgun (or probably any bullet based barrel)

Assets/Scripts/Augment/ProjectileController.cs
Fueredoriku: Ye, we should probably go on a debug hunt soon (tm)

Assets/Scripts/Augment/GunController.cs
toberge: FIXED THE INDEX THING

Fueredoriku (approved):
Looks great, amazing work for fixing that ancient Vfxformatter bug 🎉

PR #373, opened 2023-12-10
Overhaul toon shader

toberge:
- [x] Organize the toon shader graph a lil better :upside_down_face: 
- [x] Add specular highlights (more like activate them)
- [ ] Fix normal shenanigans
  - :warning: can be postponed as this requires updating *several* models
- [x] Add tiling for texture(s)
- [x] Add support for metallic textures
- [x] Add support for normal maps
- [x] Actually use toon shader in more materials (!)

image


Fueredoriku (approved):
Looks amazing, great work! 😮

My only feedback is that we should consider turning off specular on these blocks in grand canyon 
image
To this:
image

In a future PR we could also consider giving the specular bloom discrete steps, but I think the current bloom looks excellent and is better than what we had before👍

I'll leave an approve for you to use at your discretion!

toberge:
> My only feedback is that we should consider turning off specular on these blocks in grand canyon
Done, thanks!

Issue #345, opened 2023-11-26
Melee


Fueredoriku:
Dedicated melee button.

Could also feature augments that benefit from this (chainsaw bayonett)

Issue #329, opened 2023-11-26
Zooming-extension


Fueredoriku:
New extension for allowing players to zoom (scope) with their weapon

PR #376, opened 2023-12-19
Decouple first person and third person gun

Fueredoriku:
Major addition/rework for gun system
- First person guns and the gun other players see (display gun) are now decoupled
- Animations and effects like muzzle flash are now also decoupled
- Intensive runtime effects like trails and mesh projectiles are shared between both for the sake of efficiency 
- Add see-through effect when first person guns are inside geometry
fancyCulling
- Add improved IK-animations for gun holding
gunDecoulped

This makes it possible to add features like #345 and #329 + ironsight for all weapons, apply camera FOV and weapon FOV separately, and FINALLY fix the AimingEdgeCase quirk once an for all.

Code could be better in places like GunFactory, but would require an even larger refactoring (of ProjectileControllers), which I deem outside the scope of this PR

### DIFF
--- a/Assets/Scripts/Augment/AugmentImplementations/MowerAnimator.cs
+++ b/Assets/Scripts/Augment/AugmentImplementations/MowerAnimator.cs
@@ -12,6 +12,7 @@ public class MowerAnimator : AugmentAnimator
     private Animator handAnimator;
     [SerializeField]
     private ParticleSystem exhaustParticles;
+
     public override void OnFire(GunStats stats)
     {
         animator.SetTrigger("PistonPump");
--- a/Assets/Scripts/Augment/AugmentImplementations/MowerAnimator.cs
+++ b/Assets/Scripts/Augment/AugmentImplementations/MowerAnimator.cs
@@ -1,0 +1,30 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class MowerAnimator : AugmentAnimator
+{
+    [SerializeField]
+    private LineRenderer handString;
+    [SerializeField]
+    private Animator animator;
+    [SerializeField]
+    private Animator handAnimator;
+    [SerializeField]
+    private ParticleSystem exhaustParticles;
+    public override void OnFire(GunStats stats)
+    {
+        animator.SetTrigger("PistonPump");
+        handAnimator.SetTrigger("Pull");
+    }
+
+    public override void OnInitialize(GunStats stats)
+    {
+        handString.gameObject.SetActive(true);
+    }
+
+    public override void OnReload(GunStats stats)
+    {
+        exhaustParticles.Play();
+    }
+}
--- a/Assets/Scripts/Augment/AugmentImplementations/RevolverAnimator.cs
+++ b/Assets/Scripts/Augment/AugmentImplementations/RevolverAnimator.cs
@@ -38,7 +38,7 @@ public class RevolverAnimator : AugmentAnimator
         extension = transform.parent.gameObject.GetComponentInChildren<GunExtension>();
         if (extension)
             extension.transform.SetParent(attachmentSite, true);
-        Debug.Log("Reload called");
+
         animator.SetTrigger("Reload");
     }
 }
--- a/Assets/Scripts/Augment/AugmentImplementations/RevolverAnimator.cs
+++ b/Assets/Scripts/Augment/AugmentImplementations/RevolverAnimator.cs
@@ -19,7 +19,6 @@ public class RevolverAnimator : AugmentAnimator
     {
         animator = GetComponent<Animator>();
         var playerHand = GetComponentInChildren<PlayerHand>(includeInactive: true);
-        //playerHand.transform.SetParent(null);
         Destroy(playerHand);
         if (!animator)
             Debug.Log("Revolver Body missing Animator");
--- a/Assets/Scripts/Augment/AugmentImplementations/RevolverAnimator.cs
+++ b/Assets/Scripts/Augment/AugmentImplementations/RevolverAnimator.cs
@@ -10,16 +10,16 @@ public class RevolverAnimator : AugmentAnimator
     private Transform attachmentSite;
     private GunBarrel barrel;
     private GunExtension extension;
-    public override void OnFire(GunStats stats)
-    {
 
-    }
+    public override void OnFire(GunStats stats){}
 
     public override void OnInitialize(GunStats stats)
     {
         animator = GetComponent<Animator>();
         var playerHand = GetComponentInChildren<PlayerHand>(includeInactive: true);
         Destroy(playerHand);
+        barrel = transform.parent.gameObject.GetComponentInChildren<GunBarrel>();
+        extension = transform.parent.gameObject.GetComponentInChildren<GunExtension>();
         if (!animator)
             Debug.Log("Revolver Body missing Animator");
     }
@@ -30,11 +30,9 @@ public class RevolverAnimator : AugmentAnimator
         if (!canReload)
             return;
 
-        barrel = transform.parent.gameObject.GetComponentInChildren<GunBarrel>();
         if (barrel)
             barrel.transform.SetParent(attachmentSite, true);
 
-        extension = transform.parent.gameObject.GetComponentInChildren<GunExtension>();
         if (extension)
             extension.transform.SetParent(attachmentSite, true);
 
--- a/Assets/Scripts/Augment/AugmentImplementations/RevolverAnimator.cs
+++ b/Assets/Scripts/Augment/AugmentImplementations/RevolverAnimator.cs
@@ -19,7 +19,7 @@ public class RevolverAnimator : AugmentAnimator
     {
         animator = GetComponent<Animator>();
         var playerHand = GetComponentInChildren<PlayerHand>(includeInactive: true);
-        playerHand.transform.SetParent(null);
+        //playerHand.transform.SetParent(null);
         Destroy(playerHand);
         if (!animator)
             Debug.Log("Revolver Body missing Animator");
--- a/Assets/Scripts/Augment/AugmentImplementations/RevolverAnimator.cs
+++ b/Assets/Scripts/Augment/AugmentImplementations/RevolverAnimator.cs
@@ -1,0 +1,44 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class RevolverAnimator : AugmentAnimator
+{
+    private Animator animator;
+    private bool canReload = false;
+    [SerializeField]
+    private Transform attachmentSite;
+    private GunBarrel barrel;
+    private GunExtension extension;
+    public override void OnFire(GunStats stats)
+    {
+
+    }
+
+    public override void OnInitialize(GunStats stats)
+    {
+        animator = GetComponent<Animator>();
+        var playerHand = GetComponentInChildren<PlayerHand>(includeInactive: true);
+        playerHand.transform.SetParent(null);
+        Destroy(playerHand);
+        if (!animator)
+            Debug.Log("Revolver Body missing Animator");
+    }
+
+    public override void OnReload(GunStats stats)
+    {
+        canReload = !canReload;
+        if (!canReload)
+            return;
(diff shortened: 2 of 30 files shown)


toberge (changes requested):
image
tem lines not roight?
and could you double check that you didn't accidentally break the pan extension?

seems good otherwise :+1:

Assets/Scripts/Augment/GunFactory.cs
```diff
@@ -63,7 +110,7 @@ public static string GetGunName(Item body, Item barrel, Item extension)
     [SerializeField]
     public Item Extension;
 
-    private GunController gunController;
+    public GunController gunController { private set; get; }
```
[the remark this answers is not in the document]
Fueredoriku: I was a bit too lazy to change it first time around, but it has been refactored now 👍

Assets/Scripts/Augment/AugmentImplementations/RevolverAnimator.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        canReload = !canReload;
+        if (!canReload)
+            return;
+
+        barrel = transform.parent.gameObject.GetComponentInChildren<GunBarrel>();
+        if (barrel)
+            barrel.transform.SetParent(attachmentSite, true);
+
+        extension = transform.parent.gameObject.GetComponentInChildren<GunExtension>();
```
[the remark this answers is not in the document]
Fueredoriku: Done!

Assets/Scripts/Augment/AugmentImplementations/RevolverAnimator.cs
```diff
@@ -0,0 +1,43 @@
...
+
+public class RevolverAnimator : AugmentAnimator
+{
+    private Animator animator;
+    private bool canReload = false;
+    [SerializeField]
+    private Transform attachmentSite;
+    private GunBarrel barrel;
+    private GunExtension extension;
+    public override void OnFire(GunStats stats)
```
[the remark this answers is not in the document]
Fueredoriku: Fixed 👌

Assets/Scripts/Augment/GunFactory.cs
```diff
@@ -18,9 +18,56 @@ public static GameObject InstantiateGun(Item bodyPrefab, Item barrelPrefab, Item
...
+            firstPersonGunController.onFire += displayGun.GunController.PlayRecoil;
+
+        if (displayGun.GunController.projectile.GetType() == typeof(BulletController))
+        {
+            ((BulletController) gun.GetComponent<GunFactory>().GunController.projectile).Trail.layer = 0;
+            firstPersonGunController.onFire += ((BulletController)displayGun.GunController.projectile).PlayMuzzleFlash;
+        }
+
+        if (displayGun.GunController.projectile.GetType() == typeof(MeshProjectileController))
+            ((MeshProjectileController) gun.GetComponent<GunFactory>().GunController.projectile).Vfx.gameObject.layer = 0;
```
toberge, line 54: We don't have a way to just disable all the VFX calculations for the third-person guns?

Assets/Scripts/Control&Input/PlayerMovement.cs
```diff
@@ -249,7 +259,7 @@ private void UpdateRotation()
     {
         aimAngle += inputManager.lookInput * lookSpeed * Time.deltaTime;
         // Constrain aiming angle vertically and wrap horizontally.
-        aimAngle.y = Mathf.Clamp(aimAngle.y, -Mathf.PI / 2, Mathf.PI / 2);
+        aimAngle.y = Mathf.Clamp(aimAngle.y, -Mathf.PI / 2 + Mathf.Deg2Rad, Mathf.PI / 2 - Mathf.Deg2Rad);
```
toberge, line 262: Why's Deg2Rad here now, and as an additive term??? It seems to work though, but isn't that just because 2Π/360 is a very small value?

Assets/Shaders/ddrShader.shadergraph
toberge, line 1: Huh, what happened in this shader?

toberge (approved):
Now the string is held by the hand at least, some oddities are expected
image

What *is* an issue though is how easy it is to die to your own hats :upside_down_face: 
Just walk forwards while shooting hats, and you spondunkle and die.
I suspected that this was introduced when you upscaled the hats, if their hitbox was made larger as well, but it actually seems like this issue appeared in *this* PR, as the current dev branch doesn't have this problem.

I leave an approve here since things seem fine otherwise.

Assets/Scripts/Augment/AugmentImplementations/RevolverAnimator.cs
```diff
@@ -0,0 +1,41 @@
...
+        barrel = transform.parent.gameObject.GetComponentInChildren<GunBarrel>();
+        extension = transform.parent.gameObject.GetComponentInChildren<GunExtension>();
+        if (!animator)
+            Debug.Log("Revolver Body missing Animator");
+    }
+
+    public override void OnReload(GunStats stats)
+    {
+        canReload = !canReload;
+        if (!canReload)
```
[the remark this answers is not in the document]
Fueredoriku: Yes, it's a workaround for the revolver not actually reloading before animation is done, and the animation is done on... reload. It worked this way previously too for the visual of the HUD reloading only when the animation has finished, it's just more visible that it works that way now...

Assets/Scripts/Control&Input/PlayerMovement.cs
Fueredoriku: If we lock it at the very top hand animations tend to break due to shortest path slerping. This is a very unnoticable (in game) and computationaly inexpensive way to solve this issue

Assets/Shaders/ddrShader.shadergraph
Fueredoriku: Is now unlit, as the culled hand put so much shadows over the screen material that it would become invisible

Assets/Scripts/Augment/GunFactory.cs
Fueredoriku: This function is actually the opposite: letting all players see the first-person mesh projectiles, disabling the calculation of third person gun projectiles happens another place. This way we only have to calculate the first person projectiles as it's oone of the most expensive calls we have

Assets/Scripts/Augment/AugmentImplementations/MowerAnimator.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    [SerializeField]
+    private LineRenderer handString;
+    [SerializeField]
+    private Animator animator;
+    [SerializeField]
+    private Animator handAnimator;
+    [SerializeField]
+    private ParticleSystem exhaustParticles;
+    public override void OnFire(GunStats stats)
```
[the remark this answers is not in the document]
Fueredoriku: Will fix

Assets/Scripts/Augment/GunFactory.cs
```diff
@@ -18,9 +18,56 @@ public static GameObject InstantiateGun(Item bodyPrefab, Item barrelPrefab, Item
...
+        var firstPersonGunController = gun.GetComponent<GunFactory>().GunController;
+        var gunAnimations = displayGun.GetComponentsInChildren<AugmentAnimator>(includeInactive: true);
+        foreach (var animation in gunAnimations)
+        {
+            animation.OnInitialize(firstPersonGunController.stats);
+            firstPersonGunController.onFire += animation.OnFire;
+            firstPersonGunController.onReload += animation.OnReload;
+        }
+
+        firstPersonGunController.RightHandTarget = displayGun.GunController.RightHandTarget;
```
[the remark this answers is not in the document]
Fueredoriku: It is not the same point. "e need to explictly say to the gunController which actually has a player refference (Display gun does not!) that the inverse kinematics for hands (Which is attached to players) should follow the display gun and NOT the first person gun. This is because we only want to calculate IK once and only for where it matters, a third person view.

The IK automatically handles orientation/position afterwards

Assets/Scripts/Augment/GunFactory.cs
toberge: but... does setting layer to 0 actually disable them?

Assets/Scripts/Augment/GunFactory.cs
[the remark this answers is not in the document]
toberge: Oh okay, I forgor that we don't show right hands in first person :skull:

Assets/Scripts/Augment/BulletController.cs
```diff
@@ -19,6 +19,7 @@ public class BulletController : ProjectileController
 
     [SerializeField]
     private VisualEffect trail;
+    public GameObject Trail => trail.gameObject;
```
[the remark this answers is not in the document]
toberge: Okay

Assets/Scripts/Augment/AugmentImplementations/RevolverAnimator.cs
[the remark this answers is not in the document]
toberge: Yeah I'll probably be rewiring this along with rebasing the coil upgrade :upside_down_face:

Assets/Scripts/Control&Input/PlayerMovement.cs
toberge: Okay, but why not set it to some custom low value constant instead of using Deg2Rad, which I suspect would confuse others just as much about the intent of this slight alteration

Assets/Shaders/ddrShader.shadergraph
toberge: okok one piece of fedback: the arrow flash when you hit the target doesn't seems nearly as bright now, could you increase the intensity of that flash?

toberge:
Seems like you got the hat collision fixed, but I'm still very skeptical towards the inside-wall effect on the gun as it is *very* noticeable and also appears on players, guns and all sorts of other things:

image

I do see that this could be difficult and/or expensive to change, but we should seriously consider changing it since it goes against what the majority of shooters do.

Assets/Scripts/Augment/GunFactory.cs
Fueredoriku: No, it ensures that no players cull that mesh projectile

Assets/Shaders/ddrShader.shadergraph
Fueredoriku: Yup

Assets/Scripts/Control&Input/PlayerMovement.cs
Fueredoriku: My intention is to move it 1 degree, so I could go 
`
ROTATION_OFFSET = 1;

aimAngle.y = Mathf.Clamp(aimAngle.y, -Mathf.PI / 2 + Mathf.Deg2Rad * ROTATION_OFFSET, Mathf.PI / 2 - Mathf.Deg2Rad * ROTATION_OFFSET);
`
But that seems like very overkill considering Mathf.Deg2Rad is literally a constant value of 1 degree in radians, and we probably dont what to changer this value very often

Fueredoriku: I can write a dev comment  above the function though, if you think the readability is suffering too much atm

Fueredoriku:
>but we should seriously consider changing it since it goes against what the majority of shooters do.

Considering it currebtly looks way better than what we ha previously (clipping into meshes) I suggest we add it as is now, and then we can discuss what we do in the next meeting and/or demo with the change to get feedback

Assets/Scripts/Control&Input/PlayerMovement.cs
toberge: Aha, that sounds quite sensible actually, but I do think there should be a comment explaining what the intent is (as in, that Deg2Rad is used for offsetting by 1 degree and why offsetting is necessary)

Assets/Scripts/Augment/GunFactory.cs
toberge: Oh, I didn't notice that it is set on the _actual_ gun and not the third-person display gun :sweat_smile:

Issue #346, opened 2023-11-26
Update playerUI


Fueredoriku:
- [ ] Upgrade damage UI
- [ ] Move bullet meter to the right part of the screen
   -  Add bullet count (x of y bullets remaing) overlayed
- [x] Make health slider a bit cooler (color + sine waves with periods affected by health %)
- [x] #461
- [x] Movement feedback (speed lines)

PR #379, opened 2023-12-25
Speed lines

Fueredoriku:
Added speedlines which are affected by both velocity magnitude and direction, for #346 

This is mainly intended to give players some extra visual feedback for leaping, leap dashing, and dramatic effects (like when riding up fans and flying up fast with explosive barrels / falling fast)

### DIFF
--- a/Assets/Scripts/Gamestate/PlayerManager.cs
+++ b/Assets/Scripts/Gamestate/PlayerManager.cs
@@ -158,7 +158,10 @@ public class PlayerManager : MonoBehaviour
     {
         inputManager = playerInput;
         identity = inputManager.GetComponent<PlayerIdentity>();
-        GetComponent<PlayerMovement>().SetPlayerInput(inputManager);
+        var playerMovement = GetComponent<PlayerMovement>();
+        playerMovement.SetPlayerInput(inputManager);
+        playerMovement.onLeapContionous += EnableHudLeap;
+        playerMovement.onLeapEnd += DisableHudLeap;
         // Subscribe relevant input events
         inputManager.onFirePerformed += Fire;
         inputManager.onFireCanceled += FireEnd;
@@ -186,6 +189,12 @@ public class PlayerManager : MonoBehaviour
             //Remove the gun
             Destroy(gunController.gameObject);
         }
+        if (TryGetComponent(out PlayerMovement playerMovement))
+        {
+            playerMovement.onLeapContionous -= EnableHudLeap;
+            playerMovement.onLeapEnd -= DisableHudLeap;
+        }
+
     }
 
     private void UpdateAimTarget(GunStats stats)
@@ -212,6 +221,16 @@ public class PlayerManager : MonoBehaviour
         gunController.triggerHeld = true;
         gunController.triggerPressed = true;
         StartCoroutine(UnpressTrigger());
+    }
+
+    private void EnableHudLeap(Vector3 velocity)
+    {
+        hudController.SetSpeedLines(true, velocity);
+    }
+
+    private void DisableHudLeap(Vector3 velocity)
+    {
+        hudController.SetSpeedLines(false, velocity);
     }
 
     private void UpdateHudFire(GunStats stats)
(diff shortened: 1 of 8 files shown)


toberge (changes requested):
First impressions:

- The effect looks cool, but isn't as prominent as I expected
- It's almost invisible on the sides, and generally I think it should extend further inward so it is more effective
- It cuts off too abruptly when your speed tapers off, a smoother transition would be desirable
- The split between top/bottom of screen feels weird sometimes, like facing down while going up/down has speed lines on different parts of the screen

Fueredoriku:
I don't think the speedlines should be too prominent. The speed lines are also tied directly to the rigibody velocity, so calculating a fall off when the movement comes to a halt also looks weird (speedlines when standing completly still looks stanger imo)

Assets/Scripts/Gamestate/PlayerManager.cs
```diff
@@ -214,6 +221,11 @@ private void Fire(InputAction.CallbackContext ctx)
         StartCoroutine(UnpressTrigger());
     }
 
+    private void UpdateHudLeap(Rigidbody body)
```
toberge, line 224: Iffy name, just UpdateSpeedLines instead?

(also, it feels like the HUD should be subscribing to these events itself)

Assets/Scripts/UI/PlayerHUDController.cs
```diff
@@ -55,12 +56,26 @@ public class PlayerHUDController : MonoBehaviour
...
+    [SerializeField]
+    private Image speedLines;
+    [SerializeField]
+    private AnimationCurve speedLineEase;
+    private Material speedLinesMaterial;
+    private float oldLargeVelocity = 0f;
+    private const float lineDampeningVelocity = 11f;
+    private const float lineRemovalMultiplier = 0.8f;
+    private const float lineVelocityDampeningX = 0.25f;
+    private const float lineVelocityDampeningY = 0.1f;
```
toberge, line 68: These parameters could be serialized and documented with tooltips

Assets/Scripts/UI/PlayerHUDController.cs
toberge, line 1: This file is getting quite big now, perhaps we could move the speed line code to its own component instead of further bloating this one?

Assets/Shaders/SpeedLines.shadergraph
toberge, line 1: Could you group the nodes with some sensible naming, so this shader graph becomes a little more tidy? I sorta see what's going on here, but it's a little chaotic :sweat_smile:

Assets/Scripts/UI/PlayerHUDController.cs
```diff
@@ -74,6 +89,30 @@ void Start()
...
+        var magnitude = velocity.magnitude;
+        if (magnitude < lineDampeningVelocity)
+        {
+            if (oldLargeVelocity < lineDampeningVelocity)
+            {
+                speedLinesMaterial.SetFloat("_LineRemovalRadius", 1f);
+                return;
+            }
+
+            var lerpedMagnitude = Mathf.Lerp(oldLargeVelocity, 1f, Time.fixedDeltaTime);
```
toberge, line 103: dampenedMagnitude?

toberge (approved):
Them lines be looking much cooler & gooder now, just got sum commentus momentus

May need some tweaking to be less sensitive but keep its coolness (and perhaps keep its coolness for longer)

Assets/Scripts/UI/PlayerHUDController.cs
Fueredoriku: If I serialize them they can't be const. I'll add devcomments instead

Assets/Scripts/UI/PlayerHUDController.cs
Fueredoriku: I'll add a separate issue for refactoring PLayerHUD considering we then should consider extracting other components too.

Assets/Shaders/SpeedLines.shadergraph
Fueredoriku: Added some documentation in the graph 👍

Assets/Scripts/Gamestate/PlayerManager.cs
Fueredoriku: Updated name, I'll leave the scope thing for refactoring issue

Assets/Scripts/UI/PlayerHUDController.cs
Fueredoriku: Fixed 👍

Fueredoriku:
Added #401 for later

PR #386, opened 2023-12-29
Use toon shader on players

toberge:


Fueredoriku:
Great stuff! Could you also slap the material on the playerHand prefab?

Fueredoriku (approved):
Very nice!

PR #387, opened 2023-12-30
Toonify and clean up materials

toberge:
- Toon all the things
- Delete/replace unused/obsolete materials
- Fix normal maps in toon shader (transform from tangent to world space)
- Reorganize file structure for models and materials


toberge:
Note to self: try spawning players in empty children of the current spawnpoints

Fueredoriku (changes requested):
Should fix:
Seeker body's ammo (on the seeker body prefab itself) material is missing

For later:
We should have bloom with discrete steps at some points. The panels in grand crevice look good at close distance to medium distance, but looks less "right" with the style when looked at from a far, especially from above where the specular effect covers a larger area.  I suggest we create a separate issue for discrete bloom steps when merging this.

toberge:
> discrete bloom steps

Do note that bloom and specular highlights are *not* the same, and I think you mean discrete _specular_ steps here. Before making the specular highlights discrete, we may need to fix the normals of _several_ models in order to avoid  results that are further exaggerated by discretizing the steps. I haven't looked into discretizing the bloom effect yet, and I'm not certain that we should do so.

toberge:
Seeker material fixed

Fueredoriku:
I think we should set the AuctionToon material's smoothnes to 1.

We might consider setting smoothness to 1 on the materials of models that are currenty broken (like the antenna in cratertown). 

I think also setting the smoothness of the plates in grand canyon to about 0.9 might be better? I think my main gripe with grand canyon is that the plates looks so much darker, maybe we could just the color/brightness to compensate? 🤔

Previous (lighter material colors):
image

Current (darker colors):
image

Fueredoriku (approved):
Great stuff! grand canyon now has the best of both of worlds 🤠👌

PR #388, opened 2023-12-30
Add flying soda cans

### LINKED ISSUE #389: Spraying soda cans

toberge:
- Placed soda cans in a few spots in current maps
- Gave them hitboxes
- and the ability to swoooosh around
- with world-space particles leaving a looooong trail

Closes #389

### DIFF
--- a/Assets/Scripts/Augment/BulletModifiers/DecalModifier.cs
+++ b/Assets/Scripts/Augment/BulletModifiers/DecalModifier.cs
@@ -50,6 +50,7 @@ public class DecalModifier : MonoBehaviour, ProjectileModifier
     private void OnHit(Collider target, ref ProjectileState state)
     {
         // Soda cans are tiny and flail around, making them unsuitable for placing bullet holes on
+        // TODO Add some way of determining which *other* objects to avoid placing decals on
         if (target.TryGetComponent<SodaCan>(out var _))
             return;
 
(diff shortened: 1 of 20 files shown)


Assets/Scripts/Augment/BulletModifiers/DecalModifier.cs
```diff
@@ -49,6 +49,10 @@ public void Detach(ProjectileController projectile)
 
     private void OnHit(Collider target, ref ProjectileState state)
     {
+        // Soda cans are tiny and flail around, making them unsuitable for placing bullet holes on
+        if (target.TryGetComponent<SodaCan>(out var _))
```
Fueredoriku, line 53: Hmmm, maybe we should have some kind of way to mark certain objects as "not allowed as decal parents" in the furture? I don't think it's necessary yet ofc

Assets/Scripts/Interactables/SodaCan.cs
```diff
@@ -0,0 +1,80 @@
...
+        body.AddForce(flyingDirection * flyingForce, ForceMode.VelocityChange);
+    }
+
+    private void OnCollisionEnter(Collision collision)
+    {
+        if (!isFlying) return;
+
+        var contact = collision.GetContact(0);
+
+        Debug.Log("ENTERED");
```
Fueredoriku, line 74: Remove debug

Fueredoriku (approved):
Very fun!! 😁

I only have 2 nitpicks, so I'll leave an approve you can use at your discretion

Assets/Scripts/Augment/BulletModifiers/DecalModifier.cs
toberge: Yeah, and I think it'd also be useful to make certain objects have decals that display *only on that object*, for things that move around. It was absolutely necessary to add this check for the soda cans since otherwise they'd roll around and project bullet holes onto everything in a distractingly unrealistic manner.

Assets/Scripts/Interactables/SodaCan.cs
toberge: Done.

Issue #383, opened 2023-12-29
New settings


Fueredoriku:
- [ ] FoV
- [ ] Weapon FoV
- [ ] Keybindings
- [ ] Anti-Aliasing options (?)

PR #394, opened 2023-12-31
Fix mouse look jitter (partially) 

toberge:
Time.deltaTime should not be used for mouse delta since it already is a delta of pixels moved across the screen. Dampening mouse input by two orders of magnitude seems to work well for this input.

Background: I'm working on the game from a laptop where only the integrated graphics work, and with low framerate the current mouse input handling showed glaring faults, jittering painfully when I tried to aim (especially with ironsights).

### DIFF
--- a/Assets/Prefabs/Input/PlayerInput.prefab
+++ b/Assets/Prefabs/Input/PlayerInput.prefab
@@ -79,7 +79,7 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   playerInput: {fileID: 0}
-  mouseLookScale: 0.1
+  mouseLookScale: 0.0015
   gamepadLookScale: 1
   ZoomActive: 0
 --- !u!20 &-8626314750790423148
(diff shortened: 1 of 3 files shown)


Fueredoriku (changes requested):
Could you set the lookspeed to 0.0015 instead? After some testing I found this value to feel less "sluggish"

toberge:
> Could you set the mouse lookspeed to 0.0015 instead? After some testing I found this value to feel less "sluggish", as I find the mouselook speed here a bit too slow

Done. And by the way, I think we could benefit from using separate zoomed-in look speeds for mouse and gamepad, as the current zoomed-in sensitivity seems fine on gamepad and too sensitive with a mouse (though that may be subjective ofc ofc)

Fueredoriku (approved):
>  I think we could benefit from using separate zoomed-in look speeds for mouse and gamepad, as the current zoomed-in sensitivity seems fine on gamepad and too sensitive with a mouse

I think we can throw it in as a subtask in #383, as that would be nice for players to tune themselves too for the sake of personal prefferences

toberge:
I do notice that the jitter is still present on my end, so we may need to do some actual smoothing of the inputs here later

Issue #391, opened 2023-12-31
Nerf sodie popper


Fueredoriku:
Ideally, sodie popper only has infinite ammo with shotgun + knockback extension

- Should at the very least dampen the "jiggle"-force gained from recoil

Issue #358, opened 2023-12-03
Shotgun should have higher recoil omg


toberge:

Fueredoriku:
We just need to be careful with the "sodie popper". IMO that and shotgun should not = infinite ammo

Issue #355, opened 2023-11-29
Make the pan BIG


toberge:
gotta protect you more

perhaps make the collider larger than the pan

Issue #304, opened 2023-11-15
Adjust ammo capacity based on barrel


toberge:
Should be able to reduce it by a factor and then ceil it so it's at least 1

PR #398, opened 2024-01-03
Fix/augment rebalancing

Fueredoriku:
A rebalance and polish of every body augment and some barrels and extensions. Now they should be a lot more balanced (compared to each other). Some augments were changed a lot: Sodie popper, mower and pan. Also: magazine size is now actually a modifiable stat!

mowerAndPan

First person hands!
image

Detailed overview of tasks done:
- [x] #391 
- [x] Rework the Mower (Needs a buff)
- [x] Lower projectile speeds a bit (for both some bodies and hat barrel)
- [x] Lower ammo for shotgun
- [x] Fix visuals on some augments (like hacking wires)
- [x] Add PlayerArmRight to first person bodies
- [x] #358 
- [x] #355 
- [x] #304


Assets/Scripts/Augment/GunStats.cs
```diff
@@ -27,9 +27,13 @@ public enum FireModes
 
 
     //How many projectiles in a clip
-    public int magazineSize = 20;
-
-    public int Ammo = 20;
+    [SerializeField]
+    private ModifiableFloat magazine = new ModifiableFloat(20f);
+    public ModifiableFloat Magazine { get => magazine; }
```
toberge, line 32: Please use a more understandable name than just magazine for the floating point magazine size that we adjust, and use => without the `get` keyword since there's no setter.

Assets/Scripts/Augment/GunStats.cs
```diff
@@ -27,9 +27,13 @@ public enum FireModes
...
+    [HideInInspector]
+    public int MagazineSize { get => Mathf.Max(1, Mathf.RoundToInt(magazine.Value())); }
```
toberge, line 34: Same issue with this property getter (though the name is fine)

Assets/Scripts/Control&Input/PlayerMovement.cs
```diff
@@ -92,6 +93,7 @@ public class PlayerMovement : MonoBehaviour
 
     [SerializeField, ReadOnly]
     private PlayerState state = PlayerState.GROUNDED;
+    public bool StateIsAir => state == PlayerState.IN_AIR;
```
toberge, line 96: perhaps just IsInAir?

Assets/Scripts/Augment/AugmentImplementations/Pan.cs
```diff
@@ -0,0 +1,48 @@
...
+        playerMovement = gunController.Player.GetComponent<PlayerMovement>();
+
+    }
+
+    private void TryTrickJump(InputAction.CallbackContext ctx)
+    {
+        if (playerMovement)
+            if (playerMovement.StateIsAir)
+                if (Mathf.Abs(playerMovement.Body.velocity.y) < 0.01f)
+                    if (playerMovement.Body.velocity.magnitude > 1f)
```
toberge, line 37: No need to chain `if`s when you can `&&` the conditions

Assets/Scripts/Augment/AugmentImplementations/Pan.cs
```diff
@@ -0,0 +1,48 @@
...
+                        playerMovement.Body.AddForce(Vector3.up * skateJumpForce, ForceMode.VelocityChange);
+    }
+
+    private void OnDestroy()
+    {
+        if (gunController)
+            if (gunController.Player)
```
toberge, line 44: same issue here

toberge (changes requested):
- Nice solution for mag size :+1: 
- Finally the pan is more useful aaaaaa
- It is very hard to overheat the lawn mower with ""low-firerate"" barrels like the coil (yeah) and shotgun barrels, perhaps the arrow speed should be lowered a tad more on lower firerate?

And something's off with the left hand that pulls the string of the lawn mower (also could we not call it the lawn mower)
image

I'll test some more later but this is what I found for now

Assets/Scripts/Control&Input/PlayerMovement.cs
Fueredoriku: Name collision with the IsInAir method

Assets/Scripts/Augment/AugmentImplementations/Pan.cs
Fueredoriku: I know, this is style prefference.

Assets/Scripts/Augment/GunStats.cs
Fueredoriku: Was following the style of the other stats, and I refuse to have different conventions in the same class. I can change ALL stats though, which I will instead.

Fueredoriku: I also think naming it magazine is a good idea for the following reasons:

- The base body stats explicitely need an amount. While it *can* be a float, we actually want users to think it's an int anyway.
- Editing "Magazine" with multipliers in barrels and extensions doesn't feel counterintuitive considering that's how we already adjust things like "projectilespershot", which really also could be considered an int from name, but isn't.

Renaming it to something like magazineSize just seem superfluous in my opinion.

I am open to name suggestions though

Assets/Scripts/Augment/AugmentImplementations/Pan.cs
Fueredoriku: Personal style prefference under this condition, we have no convention on this

Fueredoriku:
> The cannon barrel's bubbles seem to ignore the pan, bouncing into it and exploding a few times before getting through to the player, dealing explosion damage several times

This has to do with what really is a separate issue, being cannon explosions seemingly doing damage more than once, see exmple below
image
This a whole other separate issue I don't want to delve into considering I'm already fixing 5 issues in this PR alone...

> Both the cannon and shotgun barrel still reload the sodie popper indefinitely through its knockback when you stand still 

Yes, that is indeed a combo I left in, though I did consider it. My main consideration wasn't to avoid 2 part combos that give infinite ammo under specific circumstances. It was to avoid flying infinitely unless you have specific 3 part combos, which you can't with sodie + shotgun alone after the stat changes I made.

> And something's off with the left hand that pulls the string of the lawn mower

I changed the animation, in a way I think is better (respects lower arm plane rotation constraint (You know, the disk in the elbow doesn't clip into itself)) which also avoids players seeing their shoulder which they could before when playing with only 1 screen.

Assets/Scripts/Augment/AugmentImplementations/Pan.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    void Start()
+    {
+        gunController = transform.parent.GetComponent<GunController>();
+        if (!gunController)
+            return;
+        if (!gunController.Player)
+        {
+            gameObject.transform.localScale = new Vector3(2f, 2f, 2f);
+            var hitBoxLayer = 3;
```
toberge, line 22: const int if you wanna bother :)))))

Assets/Scripts/Control&Input/PlayerMovement.cs
toberge: Ah. My thought here was just that we might want to keep state internals out and focus on the fact that this property just tells us whether the player is in the air... But this doesn't matter _that_ much, so do whatever.

Assets/Scripts/Augment/GunStats.cs
toberge: Thought of something like RawMagazineSize to indicate that this is _not_ the final value, but considering that this is _that_ one ScriptableObject class that we'd make a bloody mess by modifying... Yeah I get it.

toberge (approved):
Great work, and the lawn mower seems more balanced with the coil barrel now :+1: 

My last few concerns are exclusively regarding mag size balance, as I fear that
- the hat barrel would be much more viable with _increased_ mag size (at present it seems underpowered)
- the current mag size of the seeker might be a tad too low (dependent on level design though)
- the cannon barrel should reduce mag size to 3/4ths or something

but this is more than good enough for merging, so I'm okay with postponing further mag size balancing until we have tested s'more :sunglasses:

PR #403, opened 2024-01-05
Set overrideMatchManager to true by default

toberge:
This returns you to the menu when you start the game from one of the arena scenes... Without any side effects, I hope?


Fueredoriku (approved):
Good change! 👍

PR #405, opened 2024-01-06
Improve weapon building UI

### LINKED ISSUE #315: Revamp weapon selection screen
- [x] A/space to ready up, B/esc to cancel
- [x] Type name
- [x] Part name
- [ ] Central timer
- [x] Weapon carousel? Or just a scrolling list?
  - this should fix le jiggle issue
- [x] Make it possible to avoid picking an extension

toberge:
Closes #315 when done.

Has all intended functionality, but not that much visual polish.

image

### DIFF
--- a/Assets/Prefabs/Gamestate/BiddingRoundManager.prefab
+++ b/Assets/Prefabs/Gamestate/BiddingRoundManager.prefab
@@ -1,0 +1,1753 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!1 &1210347375213265828
+GameObject:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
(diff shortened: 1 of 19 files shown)


Fueredoriku (changes requested):
I have two requests:
- The meta file of ItemSelectManager seems to be missing the tag which makes it included in the project solution, thus needs to be regenerated it seems.
- I think selection wheel should for each augment should start at the newest aquired augment of that type instead of the first

Otherwise this looks great 😁👍

toberge:
>     * I think selection wheel should for each augment should start at the newest aquired augment of that type instead of the first

Solved this by making transactions actually "equip" the gun on the player

Fueredoriku (approved):
I also think we might want to wrap augments visually around too, but it's more of a visual thing so we can postpone that to later.

I'll leave an approve for you to use at your discretion 

Good job! 😁👍

PR #406, opened 2024-01-07
Add fancier barrel explosion

toberge:
explosion town

also fix layer of soda cans so they are affected by explosions too!

and this updates the look of the energy cannon projectile, but the explosion should look different so I haven't done anything with that


Fueredoriku (approved):
Looks so much nicer! I really enjoy the new toony look

toberge:
> Looks so much nicer! I really enjoy the new toony look

Thanks! I'm thinking of iterating on it later but I'll squeeze it in now

PR #407, opened 2024-01-07
Fix broken "rocket" jumping and player hand

toberge:
- Rocket jumping was broken on gamepads because mesh projectiles hit your own aim assist hitbox when you aim downwards :pensive: 
- Player hand subscribes to ondeath and does not unsubscribe properly (although it looks to be doing so?)


Fueredoriku (approved):
Player still hit themselves if they use the pan, but as this is better I'll leave an approve for you to use at your discretion

toberge:
> Player still hit themselves if they use the pan, but as this is better I'll leave an approve for you to use at your discretion

I think we need to look into this more closely later, as it doesn't seem to be as consistent as what I fixed here :sweat_smile:

PR #408, opened 2024-01-07
Split round end delegate

toberge:
Several immersion-breaking actions were done upon deciding the victor of the round, which is a problem when we also wait for a little while before proceeding. Split the onRoundEnd delegate into two (naming can be discussed).

also made the return-to-menu behaviour default w/o regard to override toggle, since the override does things we shouldn't do when we don't _want_ to override :upside_down_face: (but this only caused issues once I had made the other changes in this branch)

### DIFF
--- a/Assets/Scripts/Gamestate/PlayerFactory.cs
+++ b/Assets/Scripts/Gamestate/PlayerFactory.cs
@@ -30,6 +30,7 @@ public class PlayerFactory : MonoBehaviour
             SceneManager.LoadSceneAsync("Menu");
             return;
         }
+
         playerInputManagerController = PlayerInputManagerController.Singleton;
 
         // Enable splitscreen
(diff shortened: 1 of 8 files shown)


Assets/Scripts/Gamestate/PlayerFactory.cs
```diff
@@ -38,6 +39,7 @@ private void Awake()
 
         if (!overrideMatchManager)
             return;
+        // TODO omg what why does this boolean do both?
```
Fueredoriku, line 42: Maybe do something with this comment? 🤔

Fueredoriku (approved):
Seems to work! Have done some testing and it doesn't seem like this causes any issues, so it's good to go 👍

Assets/Scripts/Gamestate/PlayerFactory.cs
toberge: Removed, merging

PR #411, opened 2024-01-07
Make it possible to choose no specular

toberge:
actually part of this:
- [x] add bool for specular or not on toon shader
- [x] toggle/untoggle on all the materials
- [x] categorize toon shader parameters

tangentially related:
- [ ] make toon shader subgraph so it can be used in e.g. demo material
- [ ] fix metal map issue???

could fix the metallic map issue as well but I'm putting that off for now smh (could be solved with a step function multiplier for the specular that filters out specular at threshold .01 from metal map or sth)


Fueredoriku (approved):
Looks good! 😁👍

PR #414, opened 2024-01-08
Fix knockback on hitboxes without a rigidbody

toberge:
Caused issues with knockback + shotgun + shooting exploding barrels (haven't seen this issue elsewhere)


Fueredoriku (approved):
Seems like a good change

PR #425, opened 2024-01-14
Dampen speed lines faster for large speeds

toberge:
Previous version had speed lines from huge velocities dampen over several seconds, which felt too long

Closes #(oh we don't have an issue for it lol)


Fueredoriku (approved):
Looks good! 🤠

PR #431, opened 2024-01-20
Reduce death knockback on huge damage

toberge:
- Base death knockback on log(damage) instead of raw damage, to avoid ridiculous speeds (:warning: may need to be adjusted; requires feedback)
- Reduce explosion knockback (cannon still lifts more than barrel)
- Reduce rubber sniper damage to minimum 100 instead of minimum 620

### DIFF
--- a/Assets/Prefabs/GunParts/Bullets/CannonExplosion.prefab
+++ b/Assets/Prefabs/GunParts/Bullets/CannonExplosion.prefab
@@ -90,8 +90,8 @@ MonoBehaviour:
     m_PostInfinity: 2
     m_RotationOrder: 4
   radius: 6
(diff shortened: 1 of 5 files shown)


Fueredoriku (commented):
Seems like good changes! 😁
I would suggest not cutting down the ragdoll knockback this as much as here: The RubberSniper should maybe have a ragdoll knockback similar to an explosive barrel to give proper feedback to players in terms of how much damage it does I think. 

Could you make the push force from damage scaling slightly higher than here so the rubber sniper produce about the same knockback as an explosive barrel?

Fueredoriku (approved):
Very nice! 👍

PR #433, opened 2024-01-22
Augment symbols and Auction light fix

### LINKED ISSUE #397: Visual symbols for augment types
We should have a visual language where we can inform players in ayctions, training grounds and gun construction what type of augment they are looking at with a unique symbol

- [ ] Body symbol
- [ ] Barrel symbol
- [ ] Extension symbol

### LINKED ISSUE #423: Fix lighting issues in Auction scene
Light has started flickering and specular higlight should probably be turned off on auction stage model

Fueredoriku:
- Added augment symbols in auction and weapon construction, resolves #397 
- Fix #423 

For later:
- We should consider redoing the item pedestals in the auction scene and show the symbols on them too
- We should consider adding the symbols and some other explanatory graphics in training as well (I'll create an issue on merge)
- We should consider some peeple with better artistic skill to evaluate/redesign the symbols <.<


toberge (changes requested):
So first off, this should bring much needed visual clarity, and the symbols are fine (though the barrel may need adjustment, and the extension symbol might need some reconsideration)

However, the text on the weapon building screen now looks horribly misaligned:
image
I suggest making this text left-aligned instead and position it approx. to the right of the icons.

And in the auction, you only see the icons very briefly before the AIs (or other players) walk over and cover them with the description. Perhaps we could find some other place for them.

Fueredoriku:
> still needs adjustment on the bidding platforms but ah well

Yeah, the plan is to redo the pedestal model so it has a display where we can put the symbol, but I'll do that later (tm) 😅

Issue #334, opened 2023-11-26
Player hit feedback


Fueredoriku:
- Hitmarks
- Sparks from body part that was hit by a shot
- Player models get more worn down the more damage they have taken
- Electricity/smoke from players based on amount of damage taken

PR #438, opened 2024-01-24
Stylized skybox

Fueredoriku:
Add stylized version of our skybox!

Examples:

image

image

We might try to make the sun more visually interesting, adjust sky color hues by sun angle or make the clouds into actual gameobjects as discussed in the group, but I think this a fine iteration for now


ChrisKrane:
Love the ambience with the new skybox! About the sun, did we drop the "arms" such as seen in the concept?

Fueredoriku:
> Love the ambience with the new skybox! About the sun, did we drop the "arms" such as seen in the concept?

Not dropped, I just think its better to iterate on this version later with another PR as I need to prioritize working on #334

toberge (approved):
omg this looks so much better thank youuuuuuuu

Only nitpick is stretchy clouds and this fade looking a little wonky:  
image

PR #442, opened 2024-01-28
Fix music switching glitches

### LINKED ISSUE #372: Music loops incorrectly
Battle themes have a scheduled loop version that doesn't start at the right time.

See [this entry in the docs](https://docs.unity3d.com/ScriptReference/AudioSource.PlayScheduled.html) for a possible solution

So what needs to be done:

- [ ] way to switch between playing layers
- [ ] schedule for play like a little before it actually plays
- [ ] then swap once the loop should actually start
- [ ] fix whatever else is broken

toberge:
Uses PlayScheduled and AudioSettings.dspTime to set precise swap points
and respond to the time that the audio system actually uses.
Tracks are now loaded in extra layers some time before they are supposed
to play, at which point the layers are swapped.

Refactored the spawnpoint shuffling code as well.

Closes #372

Note that testing this PR requires waiting for the non-looping intro part of the track to end (takes like 1-3 minutes depending on the track) and verifying that no glitches or hiccups can be heard. Also note that Machiavellian Mayhem sounds a little odd in the transition to the looping version even with this fix, which I'll look into sometime soon :tm:

### DIFF
--- a/Assets/Prefabs/Sound/MusicTrackManager.prefab
+++ b/Assets/Prefabs/Sound/MusicTrackManager.prefab
@@ -1,5 +1,133 @@
 %YAML 1.1
 %TAG !u! tag:unity3d.com,2011:
+--- !u!1 &474921503203459017
+GameObject:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  serializedVersion: 6
+  m_Component:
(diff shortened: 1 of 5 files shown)


Fueredoriku (changes requested):
I can't really pinpoint how or exactly when it happens, but it seems some of the music layers are not disabled/changed after a match is complete and players are returned to the main menu. I can then hear backing (violins?) that seems to come from a non-menu song

Other than that this looks great 👍

toberge:
@Fueredoriku that issue should be resolved now

Fueredoriku (approved):
Looks good 👍

PR #448, opened 2024-01-29
Larger first person pan

Fueredoriku:
Now actually looks like a somewhat similar size to the third person size.

### DIFF
--- a/Assets/Scripts/Augment/AugmentImplementations/Hat/HatBarrelModel.cs
+++ b/Assets/Scripts/Augment/AugmentImplementations/Hat/HatBarrelModel.cs
@@ -47,24 +47,26 @@ public class HatBarrelModel : AugmentAnimator
     public override void OnFire(GunStats stats)
     {
         animator.SetTrigger("Fire");
-        for (int i = stats.Ammo; i < magazineSize; i++)
+        for (int i = stats.Ammo - 1; i < magazineSize; i++)
         {
             ammunition[i].SetActive(false);
         }
-        if (stats.Ammo == 0)
+        if (stats.Ammo == 1)
         {
-            ToggleBullet();
+            bullet.SetActive(false);
         }
     }
 
     // Called by animator!
-    public void ToggleBullet()
+    public void EnableBullet()
     {
-        bullet.SetActive(!bullet.activeInHierarchy);
-        if (!bullet.activeInHierarchy)
-        {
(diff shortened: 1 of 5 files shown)


Assets/Scripts/Augment/AugmentImplementations/Hat/HatBarrelModel.cs
```diff
@@ -47,24 +47,26 @@ public override void OnReload(GunStats stats)
...
-        for (int i = stats.Ammo; i < magazineSize; i++)
+        for (int i = stats.Ammo - 1; i < magazineSize; i++)
         {
             ammunition[i].SetActive(false);
         }
-        if (stats.Ammo == 0)
+        if (stats.Ammo == 1)
         {
-            ToggleBullet();
+            bullet.SetActive(false);
```
toberge, line 56: This `if` statement doesn't really do anything now, it just sets the bullet inactive right before the animation calls `EnableBullet`

Assets/Scripts/Augment/AugmentImplementations/Hat/HatBarrelModel.cs
```diff
@@ -47,24 +47,26 @@ public override void OnReload(GunStats stats)
     public override void OnFire(GunStats stats)
     {
         animator.SetTrigger("Fire");
-        for (int i = stats.Ammo; i < magazineSize; i++)
+        for (int i = stats.Ammo - 1; i < magazineSize; i++)
```
toberge, line 50: I noticed that when you've got 1 ammo left in the hat barrel, there are still two hats left:  
image
I believe this could be fixed by subtracting **2** from the ammo count in this for loop, and preventing -1 indexing.

toberge (approved):
Nice fixes!

Assets/Scripts/Augment/AugmentImplementations/Hat/HatBarrelModel.cs
Fueredoriku: I had to add a state due to timing restrictions, but have updated it now 👍

Assets/Scripts/Augment/AugmentImplementations/Hat/HatBarrelModel.cs
Fueredoriku: I had to do some additional state management, but yeah is fixed now 😅

PR #450, opened 2024-01-29
Fix gun stopping when AI is hit with hacking extension

Fueredoriku:
Found and fixed a bug where guns would stop working if an AI got shot with the hacking extension (since thay don't have a HUD)


toberge (approved):
They be shootin' after hackin' now, good job

PR #451, opened 2024-01-30
Add lower clouds to skybox, add thicker middle sky gradient

Fueredoriku:
Based on feedback, should look better for the arena circleing camera
image


toberge (approved):
Looks more interesting, but accentuates the seam:

image

We'll figure out something else eventually I suppose

PR #452, opened 2024-01-30
Match progress info

Fueredoriku:
A progress display for each player on scoreboards after every round


Assets/Scripts/Scoreboard/Scoreboard.cs
```diff
@@ -51,11 +60,14 @@ void Start()
...
     {
-        background.color = player.playerIdentity.color;
+        wantedPoster.GetComponent<Image>().color = player.playerIdentity.color;
         this.subtitle.text = subtitle;
         this.player = player;
+        progressPoster.GetComponent<Image>().color = player.playerIdentity.color;
+        progressDescription.text = player.playerIdentity.playerName;
 
         scoreboardManager.ShowNextCrime += NextStep;
+        scoreboardManager.DisplayProgress += ShowProgress;
```
toberge, line 70: The difference between Display and Show is... what, exactly? Some of these names could use a do-over.

Assets/Scripts/Scoreboard/ScoreboardManager.cs
```diff
@@ -140,8 +141,9 @@ private IEnumerator NextCrime()
         }
         else
         {
+            DisplayProgress?.Invoke();
             // Start next round
-            yield return new WaitForSeconds(newCrimeDelay);
+            yield return new WaitForSeconds(5f);
```
toberge, line 146: This has to match up with the timing in the Scoreboard class, right? Seems easily breakable but will work fine for the time being. Perhaps you could expose the time values in the scoreboards and use them here?

toberge (approved):
Works great, nice to finally see the match progress :tada:
 (though I stumbled upon some unrelated bugs while testing)

Assets/Scripts/Scoreboard/Scoreboard.cs
Fueredoriku: Has been updated 👍

Assets/Scripts/Scoreboard/ScoreboardManager.cs
Fueredoriku: I have now exposed the value and made individual scoreboards adjust their timings based on this value 👍

PR #453, opened 2024-01-30
Object pooling (for decals)

### LINKED ISSUE #375: Pool and limit decals on a per-gun basis
Each gun should probably have a max amount of decalls + pool their decals to avoid performance issues

image

toberge:
+ [x] Add generic object pool class
+ [x] Use this pool tool for decals
+ [x] Prevent decals from being placed on players
+  ~~use unity object pool~~ friendship ended with unity

Closes #375

### DIFF
--- a/Assets/Scripts/Utils/ObjectPool.cs
+++ b/Assets/Scripts/Utils/ObjectPool.cs
@@ -1,0 +1,38 @@
+using System.Collections.Generic;
+using UnityEngine;
+
+public class ObjectPool<T> where T : MonoBehaviour
+{
+    private T template;
+    private List<T> pool;
+    private int nextIndex = 0;
+    private int capacity;
+
+    public ObjectPool(T template, int capacity = 70)
+    {
+        this.template = template;
+        this.capacity = capacity;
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Utils/ObjectPool.cs
```diff
@@ -0,0 +1,38 @@
...
+    private int capacity;
+
+    public ObjectPool(T template, int capacity = 70)
+    {
+        this.template = template;
+        this.capacity = capacity;
+        pool = new List<T>(capacity);
+    }
+
+    public T Get()
```
Fueredoriku, line 18: It would be really funny to name this method `Surface` to go with the pooling name, but I Understand why we would opt for a name that easier to parse😅

toberge: Thought of `Pull` myself, but I felt we should go for something braindead

toberge:
As I'm dissatisfied with Unity's implementation, I think we could just go with this

Fueredoriku (approved):
Great stuff 🤠👍

PR #455, opened 2024-01-31
Fix gamebreaking bidding AI bug

toberge:
Null reference on the currentDestination in the BiddingAI script


Fueredoriku (approved):
Good change 😅👍

PR #460, opened 2024-02-02
Weapon construction visual upgrade

Fueredoriku:
- Add rotating wrap for item select
- Add icon for no extension
- Add gunstatUI always comparing to the starter weapon instead of previous weapon


Assets/Scripts/Control&Input/ItemSelectSlot.cs
```diff
@@ -52,55 +60,80 @@ public void SetItems(List<Item> items, Item initialItem)
...
+    }
+
+    private IEnumerator WaitAndSetWeapons()
+    {
+        yield return new WaitForSeconds(0.1f);
+        ChangeItemDisplayed(selectedIndex);
     }
 
-    public void Select()
+    public void Select(out Item item)
```
toberge, line 106: I'm curious why you chose to use an out parameter instead of changing the return type... And why couldn't you just use the SelectedItem property directly? There's no reason for avoiding so that I can discern

Assets/Scripts/Control&Input/ItemSelectSlot.cs
```diff
@@ -110,15 +143,24 @@ private void ChangeItemDisplayed(int nextIndex, bool animate = true)
...
-        else
+    }
+
+    public static int Mod(float a, float b)
+    {
+        float c = a % b;
+        if ((c < 0 && b > 0) || (c > 0 && b < 0))
         {
-            itemHolder.localPosition = position;
+            c += b;
```
toberge, line 162: No need to overcomplicate this, simply add `b` to `a` to avoid dropping below 0 in the first place. This method should be removed altogether, or at least turned into an extension method.

Assets/Scripts/Control&Input/ItemSelectSlot.cs
```diff
@@ -52,55 +60,80 @@ public void SetItems(List<Item> items, Item initialItem)
...
     public void Deselect()
     {
         background.color = deselectedColor;
     }
 
-    public void Previous()
+    public void Previous(out Item item)
     {
         // +count to avoid negative numbers from modulo (C# is an oddball here)
         ChangeItemDisplayed((selectedIndex - 1 + items.Count) % items.Count);
```
toberge, line 120: See this line for how I've solved it in *this very file*

Assets/Scripts/Control&Input/ItemSelectSlot.cs
```diff
@@ -52,55 +60,80 @@ public void SetItems(List<Item> items, Item initialItem)
...
+        LeanTween.moveLocal(tileHolder, itemSlots[5].transform.localPosition, .2f)
```
toberge, line 121: Um... why is this here and not part of the `for` loop in ChangeItemDisplayed?

Assets/Scripts/UI/PlayerStatUI.cs
```diff
@@ -86,7 +70,7 @@ void OnDestroy()
 
     private void OnInventoryChange(Item item)
     {
-        SetBaseGunStats(GunFactory.GetGunStats(playerIdentity.Body, playerIdentity.Barrel, playerIdentity.Extension));
+        SetBaseGunStats(GunFactory.GetGunStats(StaticInfo.Singleton.StartingBody, StaticInfo.Singleton.StartingBarrel, StaticInfo.Singleton.StartingExtension));
     }
 
     private void OnBiddingPlatformChange(BiddingPlatform platform)
```
toberge, line 76: Hm, we don't really need this one anymore, right?

toberge (approved):
Ah, now we have something more useful that another view of the gun :tada: 

The ready text and the no-extension-icon perform a sinful ritual that angers the UI gods :triumph:  
image
but I can fix that myself, it's okay

Left some comments but works well otherwise

Assets/Scripts/Control&Input/ItemSelectSlot.cs
Fueredoriku: It was to avoid convoluted nullable annotation in the return type. My original thought was that there might be cases where one isn't interested in the out variable either, but currently there is no example of that left

toberge: But still, why not use the SelectedItem property?

Fueredoriku: I'm going to change it to using that property instead, I just forgot to change it since I removed all places where the optional out was relevant 😎

Assets/Scripts/Control&Input/ItemSelectSlot.cs
Fueredoriku: I can change it to an extension method, an actual modulo operator seems like a useful thing to have instead of only a remainder operator (Thanks c# 😅)

Assets/Scripts/UI/PlayerStatUI.cs
Fueredoriku: Nope, I can remove the depreceated code 👍

Assets/Scripts/Control&Input/ItemSelectSlot.cs
Fueredoriku: Because I didn't want to animate it on the initial select / not need to the modulo difference for which direction we should move the slots. I also felt like it was slightly easier to read the code like this compared to potentially sending optional bools into the method for deciding direction and wheter to animate or not (which is my prefference, but I concede it might be non-convential in this case)

Fueredoriku: I'm gonna leave this as is for now, feel free to refactor later if you still want 😅

PR #463, opened 2024-02-04
Feature/new player select menu

Fueredoriku:


Assets/Scripts/PlayerSelect/PlayerSelectManager.cs
```diff
@@ -0,0 +1,159 @@
...
+
+        // If it doesn't, create new nametag based on player name
+        if (!nameExists)
+        {
+            Vector3 playerPosition = new Vector3(player.transform.position.x, 2, player.transform.position.z);
+
+            TMP_Text name = nameTags[playerID];
+            name.GetComponent<Transform>().position = playerPosition; // Sets the nametag position over the head of the player model
+            name.text = playerName; // Sets the text of the nametag
+            playerNames.Add(name); // Adds TMP_text object to list for monitoring
```
toberge, line 83: Some of these initialization shenanigans could be accomplished by creating a prefab with the player model and its text in the correct places... unless I'm mistaken :thinking:  (I'm **not** asking you to change this now)

Assets/Scripts/PlayerSelect/PlayerSelectManager.cs
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Starts or stops the playerselect animations coroutine. Called by "Start" and "Back" buttons in menu.
+    /// </summary>
+    /// <param name="play"></param>
+    public void controlAnimations(bool play)
```
toberge, line 91: Why not have two methods to Play() and Pause() separately?

Assets/Scripts/PlayerSelect/PlayerSelectManager.cs
```diff
@@ -0,0 +1,159 @@
...
+            string randomTrigger = "";
+            if (randomAnimatorNumber == playerInputManagerController.playerInputs.Count - 1)
+            {
+                randomTrigger = randomAnimator.GetParameter(Random.Range(2, randomAnimator.parameterCount)).name; // Choose a random trigger to set, excluding CardPeek
+            }
+            else if(randomAnimatorNumber == 0)
+            {
+                while(randomTrigger != "CardPeekReaction")
+                {
+                    randomTrigger = randomAnimator.GetParameter(Random.Range(0, randomAnimator.parameterCount)).name; // Choose a random trigger to set, excluding CardPeekReaction
```
toberge, line 127: Can't you store the parameter names upon creation and then filter out the unwanted reaction name here? Using a `while` with a random condition is a little scary, though the probability for a long time with one outcome is miniscule.

And with a list of names, you would be able to use the `RandomElement()` extension method we've prepared for this kinda thing.

...oh, and then I noticed that you *know* that CardPeekReaction is the last parameter already (from line 145), perhaps you could just replicate that line here instead?

Assets/Scripts/UI/MainMenu/MainMenuController.cs
```diff
@@ -23,6 +23,8 @@ public class MainMenuController : MonoBehaviour
     private GameObject defaultMenu;
     [SerializeField]
     private GalleryMenu galleryMenu;
+    [SerializeField]
+    private GameObject playerSelectModel;
```
toberge, line 27: Only relevant part is the `PlayerSelectManager`, perhaps use that type here so that the intent is more explicit (and change the name)

Assets/Scripts/PlayerSelect/PlayerEyeManager.cs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /// <summary>
+    /// Moves the eyes to the players left.
+    /// </summary>
+    /// <param name="targetPosition"></param>
+    public void lookLeft(float targetPosition)
+    {
+        eyeMaterial.SetVector("_RightEye", new Vector4(targetPosition, 0.0f));
+        eyeMaterial.SetVector("_LeftEye", new Vector4(targetPosition, 0.0f));
```
toberge, line 43: I just wanna see some interpolation here, but we can smooth these later

Assets/Scripts/PlayerSelect/PlayerSelectManager.cs
ChrisKrane: Redolved using a list of parameternames and RandomElement().

Assets/Scripts/PlayerSelect/PlayerEyeManager.cs
ChrisKrane: Was briefly discussed during a meeting on Wednesday 31.01. and deemed to not be an immediate necessity. We could definitely (and should probably) fix it later though

Assets/Scripts/PlayerSelect/PlayerSelectManager.cs
ChrisKrane: Because i have a bad habit of concentrating code in my methods. Fixed for clarity. :)

Assets/Scripts/PlayerSelect/PlayerSelectManager.cs
```diff
@@ -0,0 +1,159 @@
...
+    /// Fills in the playername and positions it.
+    /// </summary>
+    /// <param name="player"></param>
+    /// <param name="playerName"></param>
+    /// <param name="playerID"></param>
+    public void setPlayerNameTag(GameObject player, string playerName, int playerID)
+    {
+        // Check if nameTag exists already
+        bool nameExists = false;
+        for (int i = 0; i < playerNames.Count; i++)
```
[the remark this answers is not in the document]
ChrisKrane: Wasn't aware of this method, thank you! Added for practice.

Assets/Scripts/PlayerSelect/PlayerSelectManager.cs
ChrisKrane: Noted for later. Thank you!

PR #464, opened 2024-02-04
Add spiky backdrop for collectables

toberge:
Will add pop afterwards


Fueredoriku (changes requested):
The outlines look great 😁👍

I think we want to make the slightly smaller though, the smallest size currently could be the largest, and the smallest could be even smaller. this has the upside of avoiding some overlap issue like this:
image

Also, it seems that ammoboxes don't have the outline effect after they respawn

Fueredoriku (changes requested):
One last edge case I have found:

image

I think the effect needs to be in the ammobox layer, as the ammobox backdrop is currently viewable by people who do not have the seeker

Otherwise this seems good, both in terms of code and the effect 👍

PR #466, opened 2024-02-04
Add outline to death and victory texts

toberge:


Fueredoriku (approved):
Good change! Hopefully we'll be able to change every text instance to have an outline sooner or later 😅

PR #467, opened 2024-02-04
Add chip popup to HUD

### LINKED ISSUE #461: Give players "popup" when obtaining chips

toberge:
- [x] Textbox in correct position
- [x] Actual chip
- [x] Show updated chip amount upon collection
- [x] Animate (slide in, slide out)
- [x] Prevent chip popup on round end (!!!)
  - This sorta ties into the issue with round data not being frozen after the winner is determined. I suggest that we solve that stuff separately.
- [x] Anchor left/right if only one player (adjust anchormin/max) 

Closes #461

### DIFF
--- a/Assets/Materials/Collectables/UnlitChipMaterial.mat
+++ b/Assets/Materials/Collectables/UnlitChipMaterial.mat
@@ -1,0 +1,157 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!21 &2100000
+Material:
+  serializedVersion: 8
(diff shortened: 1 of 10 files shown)


Fueredoriku (changes requested):
There seems to be two issues:
image
The chip popup should probably be a tad bit lower, single player view produce overlap between the chips and timer

There also seem to be an issue where non-AI players are no longer able to bid on itemplatforms

Fueredoriku (approved):
Looks great! 😁👍

PR #468, opened 2024-02-05
Fix soda can trail continuing after explosion

toberge:
Turned it into a state machine to make it easier to reason about


Fueredoriku (approved):
Works like expected 👍

PR #470, opened 2024-02-05
Turn toon shader into subgraph

toberge:
image

- [x] Turned toon shader into subgraph
- [x] Used it in player eyes and demo shader (:warning: lemme know if I missed something)
- [x] Replaced old demo shader texture (it was the test texture in Source Engine)
- [x] Refurbished the old demo arena (pure nostalgia)


Fueredoriku (changes requested):
Looks pretty good, especially the glowing player eyes, but it seems we no longer have shadows, so that'll probably need to be fixed before merge

Other that that I'm happy we can finally toon shade the jigglemeshes! 🤠

toberge:
Fixed the shadow issue and toonified the jiggle materials

Fueredoriku (approved):
Great stuff!! 😁👌

I'll leave an approve and a small request:
Could you set the texture intensity on the rubber sniper to 3? 
from this:
image

to this:
image

PR #471, opened 2024-02-05
Fix auction crashing bug

Fueredoriku:
Fix bug where ais were obligatory to not crash auciton


toberge (approved):
... why isn't this the default behaviour, C# gods

PR #477, opened 2024-02-07
Texturing the player

Fueredoriku:


ChrisKrane (approved):
Looks very nice! The rust spots around the eyes/goggles and the metal spots on the rust of the hat looks a lot better than earlier. 
I can see that the back of the shoulder joint is still the base color and not colored, but i suppose this isn't critical since the model will be viewed mostly from the front.

Fueredoriku:
> back of the shoulder joint is still the base color and not colored
Yeah, it's way more of a pain to solve than one might think (due to model import issues), so I think this is good enough for now 😅

PR #478, opened 2024-02-07
Prevent players from sticking to walls

### LINKED ISSUE #479: Make player material frictionless such that players don't stick to walls.

toberge:
image

This was it. Actually. Yikes.

Edit: Wasn't just it. Needed another collider so that the feet have friction :))))))

Closes #479

### DIFF
--- a/Assets/Materials/SlipResistant.physicMaterial
+++ b/Assets/Materials/SlipResistant.physicMaterial
@@ -6,9 +6,9 @@ PhysicMaterial:
   m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInstance: {fileID: 0}
   m_PrefabAsset: {fileID: 0}
-  m_Name: Frictionless
-  dynamicFriction: 0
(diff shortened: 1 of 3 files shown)


Fueredoriku (commented):
Definetily works, but it seems the player movement has very little friction (like gliding on the ground while walking)

I think we might need to adjust either ground drag in combination with increasing ground walking force or create som kind of static drag coefficient for when players no longer apply force to slow them down faster

Fueredoriku (changes requested):
Great fix! 😁👍
My only request is that you layer the new hitbox so players can be hit
image

Fueredoriku (approved):
Movement is good, very nice that we are no longer stuck, great work! 😅

I have one final ask, could you disable the new collider before ragdoll-ification? It seems that it currently obstructs the ragdoll from falling properly/fly away when an explosion occurs.

I'll leave an approve meanwhile for you to use at your discretion 😁👍

PR #481, opened 2024-02-07
Fix more nullpointer references

toberge:
there are so many of them omg


Fueredoriku (dismissed):
Very good idea to start fixing these, I fear we might start getting client crashes sooner or later if we have too many 😬

toberge:
No point in holding onto this PR too long

Fueredoriku (approved):
Good stuff!

PR #482, opened 2024-02-07
Add Fredrik's explosion sound

toberge:
Modified to have a very short attack (it had a few hundred milliseconds of buildup, which doesn't fit an explosion).
Should be replaced by something that sounds less similar to our regular gunshots, but the lack of an explosion sound has driven me insane, so this is the result.

Basically #354 but without the merge-conflicting changes to the revolver


Fueredoriku (approved):
Very nice, happy to see my paper-smacking turned to explosion sound finally in use! 😅

PR #483, opened 2024-02-09
Add working and improved grand cayon camera panning animation

Fueredoriku:
Seems something bad happened to the last grand canyon panning camera animation, so I added it back and improved it.


toberge (approved):
lookin' good

PR #484, opened 2024-02-09
Add worse AI aim with randomness, fix AI playerhands

Fueredoriku:
- Make AI aiming slightly random, with the aim improving depending on how close the target is
- Fix first person gun playerhand being visible on AIs


toberge (approved):
Seems to work quite well! Finally, the bots are no longer extremely deadly

PR #487, opened 2024-02-10
Enhancement/new player select room

ChrisKrane:
Image_Sequence_004_0000


Assets/Scripts/PlayerSelect/PlayerEyeManager.cs
```diff
@@ -13,14 +14,21 @@ public void Start()
     {
         // Find eye material
         SkinnedMeshRenderer eyeRenderer = transform.GetChild(0).GetComponent<SkinnedMeshRenderer>();
-        eyeMaterial = eyeRenderer.materials[1];
-        resetEyePosition();
+        Material[] playerMaterials = eyeRenderer.materials;
+        foreach (Material playerMaterial in playerMaterials)
+        {
+            if (playerMaterial.name == "PlayerEyesMaterial (Instance)") {
```
toberge, line 20: Could be accomplished with a .Single(m => m.name = "blah blah")

toberge (changes requested):
Looks good!

Though, where did the cards on the table go? I felt they were a nice detail.

Also, could you make the table material toon shaded now? It seems to work quite well  with the current design:  
image

Assets/Scripts/PlayerSelect/PlayerEyeManager.cs
ChrisKrane: Fixed, and re-added cards on table. I had completely overlooked them when reimporting the new model. Thank you!

Fueredoriku (approved):
looks great! 😁👍

toberge (approved):
Very nice!

PR #488, opened 2024-02-10
Improve weapon select layout and center items

toberge:
- [x] Avoid transform-scaling the stats
- [x] Adjust layout
- [x] Add radial indicator to timer
- [x] Add midpoints to all augments
- [x] Use midpoints to center augments in
  - [x] item slots
  - [x] bidding pedestals
  - [x] training mode pedestals


Assets/Scripts/Augment/Augment.cs
```diff
@@ -59,4 +63,16 @@ public static void DisableInstance(GameObject instance, AugmentType type)
                 break;
         }
     }
+
+    public static Transform Midpoint(GameObject instance, AugmentType type)
```
Fueredoriku, line 67: Hmm, could be that bodies should inherit from augment instead of this static check, but we can postpone that refactor for later

Assets/Scripts/Control&Input/ItemSelectSlot.cs
```diff
@@ -66,20 +64,32 @@ public void SetItems(List<Item> items, Item initialItem)
...
+
+            // Item is an augment
             var rotation = Quaternion.Euler(new Vector3(0, 90, -20));
             var instance = Instantiate(items[i].augment, Vector3.zero, rotation, itemHolder);
             instance.transform.ScaleAndParent(itemModelScale, itemHolder);
-            instance.transform.localPosition = itemSlots[i].transform.position;
+            Transform itemCenter;
+            if (items[i].augmentType == AugmentType.Body)
+            {
+                itemCenter = instance.GetComponent<GunBody>().midpoint;
```
Fueredoriku, line 82: Ye, I definetly think gunbody should inherit from augment, but eh, we can refactor later

Fueredoriku (approved):
Good feature! 🤠

I think we can make gunbody inherit from augment, but we can refactor that later (I planned to do so way back in #126 ), so don't bother unless you really want to for some reason

Assets/Scripts/Augment/Augment.cs
toberge: Agreed & yeah I ain't gonna do that now haha

PR #489, opened 2024-02-10
Model and refactor flying platforms

toberge:
image

Model may be improved later, it works for now.


Fueredoriku (approved):
Very cute, great stuff 😁👍

PR #490, opened 2024-02-11
Make cacti bend when shot

toberge:


Fueredoriku (approved):
Really good!! I also really enjoy that you can bend the cacti by standing on them 😅

PR #491, opened 2024-02-11
Add sound to flying soda cans

toberge:
Something's off with the 3D sound... The crux of the issue is that it's only possible to have _one_ AudioListener in a scene


toberge:
We won't be able to do 3D sound for background noises with splitscreen multiplayer, so I'll close this :cry:

PR #492, opened 2024-02-11
Prevent specular highlights on shadow materials

toberge:


Fueredoriku (approved):
Seems like a good change 😅

PR #494, opened 2024-02-12
Add simple handle animation to weapon select, fix ready text

Fueredoriku:
Further addressing #419


toberge (approved):
omg the handle animation looks so good, thanks Fredrik xoxo

PR #496, opened 2024-02-12
Tighten player hitboxes

toberge:
- Make the box colliders for the players' physics more tailored to their lanky physique
- Change the aim assist capsule to a box with a similar shape to the resized physics colliders

May require some testing.


Fueredoriku (approved):
Seems like a good change, I haven't been able to find any issues at least 🤠

PR #499, opened 2024-02-15
Use new fonts on all the text

toberge:
aaaaa

tiny homemade styleguide living rent-free in my head and now in yours:

- Use STENCIL for
  - Menu buttons
  - Numbers
  - Metal stuff
  - Some labels
  - Mainly just numbers and beeeg stuff, it looks kinda bad at smol font size unless you bold it and use thin outline
  - Remember to bold and midline it (alignment is weird)
- Use Edmunds Distressed for
  - Distress (death, wanted posters)
  - Titles
  - Anything where Edmunds doesn't fit
- Use Edmunds for
  - Descriptions and other small text that benefits from greater clarity but thinner text (????)


Fueredoriku (approved):
Looks so much better! 😎👌

My only feedback is that with the new fonts there are now two new issues:
- Long Gun Names no longer fit in the weapon select screen
- The font scaling problem is very noticeable on the itemPlatforms

I'll leave an approve and set up a new issue for fixing the two points above 🤠

PR #502, opened 2024-02-16
Small UI upgrades

Fueredoriku:
- Add night mode for skybox in main menu
- Add player head symbol to scoreboards
- Add new symbols for augments
- Fix a knockback and tracking issue with AIs
- Slight move some components in itemselectplayer so even the long gun names can be displayed correctly


toberge (approved):
- Sunset looks quite good, but why would the clouds disappear? 🤔 
- Player heads look great, but should be inverted for better contrast (if we do change the background). Alternatively we could make the text all black, which would make the posters more similar to real wanted posters.
- Augment symbols look great and the level of detail works really well at the scale they're displayed 👍 
- The ready text seems too dark now, but it might work better with the new backdrop (why no outline though >_>)
- You probably fixed the bots, but I haven't verified 'em
- I noticed some layout change in item select, seemed fine

PR #504, opened 2024-02-17
Add grace time after ready state in item select

toberge:
Add one second where players can see their final state (when all players are ready) and press an input to edit their weapon again if they change their mind.

Also improve responsiveness for keyboard players by disabling the gamepad input delay for them.

### DIFF
--- a/Assets/Scripts/Control&Input/ItemSelectManager.cs
+++ b/Assets/Scripts/Control&Input/ItemSelectManager.cs
@@ -1,10 +1,15 @@
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using UnityEngine;
 
 public class ItemSelectManager : MonoBehaviour
 {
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Control&Input/ItemSelectManager.cs
```diff
@@ -22,11 +28,27 @@ private void Finish()
...
 
+    private IEnumerator WaitAndFinish()
+    {
+        yield return new WaitForSeconds(graceTime);
+        Finish();
+    }
+
+    private void OnNotReady(ItemSelectMenu menu)
+    {
+        Debug.Log("Notified that we no longer ready");
```
Fueredoriku, line 39: Maybe remove debug

Fueredoriku (approved):
Seems like a good quality of life change, haven't been able to test it with multiple players yet, but seems to work well 😁👍

PR #506, opened 2024-02-17
Add credits

toberge:
- needs thank you section
- mentioned Olivia in visual design since she drew concept for the auction scene at least, dunno full name & should prolly ask


karofmah (approved):
Looks great, good job!

PR #507, opened 2024-02-17
Feature/loading screen

karofmah:
loading screen


Assets/Scripts/Gamestate/MatchController.cs
```diff
@@ -167,11 +170,17 @@ public void StartNextBidding()
         PlayerInputManagerController.Singleton.ChangeInputMaps("Bidding");
         MusicTrackManager.Singleton.SwitchTo(MusicType.BIDDING);
         onBiddingStart?.Invoke();
-        // TODO: Add Destroy on match win
+        // TODO: Add Destroy on match win   
+        //playerInputManagerController.RemoveListeners();
```
toberge, line 174: Why's this commented out?

Assets/Scripts/Gamestate/MatchController.cs
```diff
@@ -167,11 +170,17 @@ public void StartNextBidding()
...
+        StartCoroutine(ShowLoadingScreen());
+    }   
+    private IEnumerator ShowLoadingScreen(){
+        loadingScreen.SetActive(true);
+        yield return new WaitForSeconds(6);
```
toberge, line 179: The 6 here should be configurable, and perhaps sit on the loading screen class itself.

Assets/Scripts/LoadingScreen.cs
```diff
@@ -0,0 +1,58 @@
...
+    private float incrementTimer = 360f;
+    private string currentTip;
+
+    private string dash = "A clever trick awaits: crouch, jump, then spring off any surface for a speedy dash to victory";
+    private string leap = "You can spring into action by crouching and then leaping in any direction with a jump";
+    private string weaponCombinations = " With more than 250 weapon combinations available, you've got endless strategies to explore"; 
+    private string skate = "You can get creative and use the pan as a makeshift skateboard for some unexpected moves";
+    private string auctionContest = "Bidding strategically to throw off your opponents can be just as effective as going for the augment yourself";
+    private string explosiveBarrels = "Leveraging explosive barrels can help you seize the high ground advantage";
+    private string saveChips = "Saving chips could be the secret to dominating the auction later on";
```
toberge, line 27: These tips could be a serialized array that we edit in Unity.

Assets/Scripts/LoadingScreen.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    private List<string> tips; 
+
+    private float rotateSpeed = 60;
+    void Start()
+    {   
+        tips = new List<string>{dash, leap, weaponCombinations, skate, auctionContest, explosiveBarrels, saveChips}; 
+        int randomIndex = Random.Range(0, tips.Count);
```
toberge, line 35: We have the `.RandomElement()` extension method for this purpose.

toberge (changes requested):
The loading screen looks really nice, and the tips seem useful!

However, I have a few concerns:
- There's no loading screen after weapon building, which is when we have the longest load time. At present, we really need a loading screen there to avoid the awkward behavior where the timer still ticks down after the weapons have been set in stone (all players ready). I can wire it up with the item select if you don't want to, but there's a TODO comment in `ItemSelectManager` where the loading screen should be started.
- The progress indicator looks good, but perhaps we shouldn't make the yellow progress fill rotate along with the revolver chamber, to make the progress more visible.

Assets/Scripts/LoadingScreen.cs
```diff
@@ -0,0 +1,58 @@
...
+    private GameObject LoadingBar;
+
+    [SerializeField]
+    private TMP_Text tipsText;
+
+    private float incrementTimer = 360f;
+    private string currentTip;
+
+    private string dash = "A clever trick awaits: crouch, jump, then spring off any surface for a speedy dash to victory";
+    private string leap = "You can spring into action by crouching and then leaping in any direction with a jump";
```
toberge, line 22: Perhaps "You can jump farther by crouching and then jumping in any direction"

Assets/Scripts/LoadingScreen.cs
```diff
@@ -0,0 +1,58 @@
...
+    [SerializeField]
+    private GameObject LoadingBar;
+
+    [SerializeField]
+    private TMP_Text tipsText;
+
+    private float incrementTimer = 360f;
+    private string currentTip;
+
+    private string dash = "A clever trick awaits: crouch, jump, then spring off any surface for a speedy dash to victory";
```
toberge, line 21: Perhaps "Crouch, jump, then jump again when you hit a surface to gain speed"

Assets/Scripts/LoadingScreen.cs
```diff
@@ -0,0 +1,58 @@
...
+    private string leap = "You can spring into action by crouching and then leaping in any direction with a jump";
+    private string weaponCombinations = " With more than 250 weapon combinations available, you've got endless strategies to explore";
```
toberge, line 23: > 250

> endless

something's fishy in these waters

Assets/Scripts/LoadingScreen.cs
```diff
@@ -0,0 +1,58 @@
...
+    [SerializeField]
+    private TMP_Text tipsText;
+
+    private float incrementTimer = 360f;
+    private string currentTip;
+
+    private string dash = "A clever trick awaits: crouch, jump, then spring off any surface for a speedy dash to victory";
+    private string leap = "You can spring into action by crouching and then leaping in any direction with a jump";
+    private string weaponCombinations = " With more than 250 weapon combinations available, you've got endless strategies to explore"; 
+    private string skate = "You can get creative and use the pan as a makeshift skateboard for some unexpected moves";
```
toberge, line 24: Perhaps remove "get creative"

Assets/Scripts/LoadingScreen.cs
```diff
@@ -0,0 +1,58 @@
...
+    private string auctionContest = "Bidding strategically to throw off your opponents can be just as effective as going for the augment yourself";
```
toberge, line 25: "going for" -> "aiming for"?

Assets/Scripts/LoadingScreen.cs
```diff
@@ -0,0 +1,58 @@
...
+    private string explosiveBarrels = "Leveraging explosive barrels can help you seize the high ground advantage";
```
toberge, line 26: Perhaps remove "advantage"

karofmah:
> The loading screen looks really nice, and the tips seem useful!
> 
> However, I have a few concerns:
> 
> * There's no loading screen after weapon building, which is when we have the longest load time. At present, we really need a loading screen there to avoid the awkward behavior where the timer still ticks down after the weapons have been set in stone (all players ready). I can wire it up with the item select if you don't want to, but there's a TODO comment in `ItemSelectManager` where the loading screen should be started.
> * The progress indicator looks good, but perhaps we shouldn't make the yellow progress fill rotate along with the revolver chamber, to make the progress more visible.
> * We should use one of the new fonts here, I'm thinking Edmunds Distressed.

* I haven't implemented this on purpose until now, to prevent conflicts. I will definitely work on this as soon as possible.
* I agree, and I have already thought about this. However, i'm unsure how to implemented this. Rotating only the revolver chamber will cause an overflow of the yellow progress. Do you have any suggestions?
image

* Done, i will push the changes as soon as possible

toberge:
> > * There's no loading screen after weapon building, which is when we have the longest load time. At present, we really need a loading screen there to avoid the awkward behavior where the timer still ticks down after the weapons have been set in stone (all players ready). I can wire it up with the item select if you don't want to, but there's a TODO comment in `ItemSelectManager` where the loading screen should be started.
>     * I haven't implemented this on purpose until now, to prevent conflicts. I will definitely work on this as soon as possible.

That's okay.

> > * The progress indicator looks good, but perhaps we shouldn't make the yellow progress fill rotate along with the revolver chamber, to make the progress more visible.
>     * I agree, and I have already thought about this. However, i'm unsure how to implemented this. Rotating only the revolver chamber will cause an overflow of the yellow progress. Do you have any suggestions?

We can deal with this later. I'd suggest avoiding parenting one to the other, though.

karofmah:
TODO:

Add controller setup and bidding instructions at the start of a match

Assets/Scripts/Control&Input/ItemSelectManager.cs
```diff
@@ -24,7 +29,13 @@ public void StartTrackingMenus()
...
     private void Finish()
     {
-        // TODO Fade to loading screen
+        StartCoroutine(LoadAndChangeScene());
+    }
+
+    private IEnumerator LoadAndChangeScene(){
+
+        loadingScreen.SetActive(true);
+        yield return new WaitForSeconds(delayDuration);
```
toberge, line 38: Perhaps the delay should be placed after the loading (which is async) has started?

Fueredoriku: FYI: 
- The method called here in matchcontroller doesn't load the the sene async (Even though it now should)

- If it did and you want the delay after such a load, one would have to write a hole block of code to stop Unity from automatically not send us to the next scene immediately.
From the Unity docs:
`
//Begin to load the [Scene](https://docs.unity3d.com/ScriptReference/SceneManagement.Scene.html) you specify
        [AsyncOperation](https://docs.unity3d.com/ScriptReference/AsyncOperation.html) asyncOperation = [SceneManager.LoadSceneAsync](https://docs.unity3d.com/ScriptReference/SceneManagement.SceneManager.LoadSceneAsync.html)("[Scene](https://docs.unity3d.com/ScriptReference/SceneManagement.Scene.html)3");
        //Don't let the Scene activate until you allow it to
        asyncOperation.allowSceneActivation = false;
`

We can use the delay to toggle that property () to guarantee atleast 3 seconds have been spent in loading no matter the load times

toberge: We'll leave this for later then

toberge (approved):
Looks good! We'll have to make some adjustments afterwards, but nice to have it in place :+1:

PR #508, opened 2024-02-18
Toon shade tumbleweeds

toberge:
yeah, they weren't. Only non-toon remaining is coil barrel (don't toon shade the coils themselves btw, we don't have emissive toon yet)

also make the script for the tumbleweed in the main menu more generic ig


Fueredoriku (approved):
Very weird that we didn't notice this earlier 😅

PR #511, opened 2024-02-18
Trigger on-hit effects through explosions

toberge:
This is a jank fix for a jank problem


Fueredoriku (approved):
A bit jank, but seems to work fine 😅

We could use a hashset for hits instead of the list of tuples, but I think the runtime costs are more or less going to be the same in practice, so I don't think it matters either way

Great to have this actually working now 🤠

toberge:
> We could use a hashset for hits instead of the list of tuples, but I think the runtime costs are more or less going to be the same in practice, so I don't think it matters either way

Considering we already loop through all colliders, this shouldn't be much of an extra cost. The explosion controller uses a hashset for its check anyways.

PR #513, opened 2024-02-19
Add color and % text to health HUD

Fueredoriku:
Also:
- Tweak scale slightly
 - Add fillface for scoreboard


toberge (changes requested):
Transitions look very nice! However, could you change to the STENCIL font (bold, midline) that is used for other numbers? And the text shouldn't cover that much space, perhaps.

toberge (approved):
Unity crashed but it looked good soooooooo let's get it in

PR #514, opened 2024-02-19
Fix regression in item select input code

toberge:
Missed one of the four directions and left some old code there that messed up the gamepad delay.


Fueredoriku (approved):
Seems good! 👍

PR #515, opened 2024-02-19
Add tutorial signs and improve educational value of training grounds

toberge:
- [x] how move
- [x] how shoot
- [x] how zoom
- [x] how jump
- [x] how crouch
- [x] how leap
- [x] how leapdeash
- [x] what chip
- [x] explosive barrel (propulsion?)
- [x] le fan (kanskje ikke så nødvendig) (gjorde noe basic)
- [x] how equip weapon here
- [x] place signs at corresponding locations and reorganize training mode to suit them
- [x] Exit sign(s)?


Fueredoriku (approved):
Looks great! 😁👍

We could consider moving training up to the landingMenu instead of level select now, but I'll just leave an approve so you can decide what you want to do

PR #521, opened 2024-02-20
Fix player join, add press button to join to character select

Fueredoriku:
- Fix players being unable to join game after a match
- Fix multiple event systems error (finally 😅)
- Add "Press button to join!" to character select


Assets/Scenes/Menu.unity
toberge, line 1: I just have one suggestions: Use the Edmunds Distressed font for the join message

toberge (approved):
Unable to verify joining with multiple controllers right now, but these seem like really great fixes.

Assets/Scenes/Menu.unity
Fueredoriku: Will do! 👍

PR #522, opened 2024-02-21
Laser barrel

Fueredoriku:
New barrel augment made by @SondreHus!

Changes include:
- Add lazer barrel!
- Fix riccochet behaviour (no longer bounces many times per surface before getting reflected)
- Add more bulging to sodie popper animation

image


toberge (approved):
Works quite well (now that I fixed it :smiling_imp:)

PR #524, opened 2024-02-21
Fix/player stuck on edge

toberge:


toberge (commented):
I'm still getting the same bug, just less severe. Try checking with the small+medium boxes in training mode.

ChrisKrane:
> I'm still getting the same bug, just less severe. Try checking with the small+medium boxes in training mode.

Should be properly fixed now. Also added Ledge detection.

Fueredoriku (approved):
Very, VERY good work @ChrisKrane! Looks great 😁👌

PR #525, opened 2024-02-21
Add keybinds and auction instructions to loading screens

toberge:


Fueredoriku (approved):
Just one small request, could we rewrite the bidding tutorial message to state:
"Bid using [...] **on a platform** to win weapon parts[...]"

After my experience using that specific phrasing helps a lot during demos.

Everything else seems good, so I'll leave an approve for you to use at your discretion 👍

PR #527, opened 2024-02-22
Make bullets pass through invisble boundaries

toberge:
Fixes the excessive weirdness introduced by the lazur barrel. Made some props static as well.


Fueredoriku (approved):
Looks good 🤠

PR #528, opened 2024-02-22
Fix training mode health overflow

toberge:
- Make it less ridiculously large
- Clamp between 0 and 100 when displaying


Fueredoriku (approved):
Thanks for fixing 👌

PR #526, opened 2024-02-21
Add the new soundFXs to the game!

Fueredoriku:
Add soundFX for:

Augments 
- revolver
- fire 
- shotgun
- seeker 
- coil
- lazur
- solar
- sodie popper
- hacking
- DDR
- Mower

Auction
- Auctioneer
- ItemPlatforms

UI
- MainMenu buttons
- Weapon select

PR #529, opened 2024-02-22
Add spam popup sound effect

toberge:
sounded better in practice


Fueredoriku:
Added in #526 instead

PR #534, opened 2024-02-23
Add cinematic intro (thank you Cristopher)

toberge:


toberge:
Unity's video player doesn't manage to start _immediately_ so I had to display the first frame on top of the menu scene and switch to the actual video once it started -_-)

ChrisKrane:
Don't know if it was decided not to fix this, but the color of the beam still changes from yellow to blue. Other than that: amazing work!!

toberge:
> Don't know if it was decided not to fix this, but the color of the beam still changes from yellow to blue. Other than that: amazing work!!

Video isn't updated here yet

Fueredoriku (approved):
Very nice! 🤠

My only piece of feedback is that we could conisder starting the menu theme early:
When we start panning down from the sun in the last scene for a "smoother" transition

Given that the video still is WIP I think we can consider doing so if still relevant later though 👍

toberge:
> My only piece of feedback is that we could conisder starting the menu theme early: When we start panning down from the sun in the last scene for a "smoother" transition

Agreed, I'll do that in a separate PR since it requires some extra logic

PR #535, opened 2024-02-23
Add new textures to weapon select

Fueredoriku:


toberge (changes requested):
I like the
- coin slot (which makes this look much more like a slot machine)
- attempt at making the UI background look like it lies on top of the gray background
- and the background on the item slots

but

- the absolute lack of color
- the overwhelming _grayness_
- the way the plaques don't have a different color than the background (they don't stand out except for the shadown, which doesn't do the job entirely)
- the fact that the camera background color is just slightly different from the UI background color
- the lack of plaques behind the three type headers (the rusty metal from the previous version looked really great there)
- the lack of contrast between the text and the background for the label triplets
- and the almost absolute lack of detail (except for some scratches that are hard to notice

makes this version look way less polished.

At present I think it would be much better to just use the previous version, perhaps with discretized colors so it fits our art style. I could get behind this new version if there's more color (in sensible ways) and some adjustments that make

toberge (approved):
Now it looks quite good, good job both of ye!

PR #541, opened 2024-02-23
Fade in menu theme in intro video

toberge:
- Fade in menu theme when the sun appears in the intro video
- Reset the sun movement when the video ends so that the menu looks approximately correct


Fueredoriku (approved):
Great stuff!

My only feedback is that:
- The skip text shows up on player joined if the first input is after the movie has played
- The sun should be set to its initial position no matter when the movie ends (either if skipped or finished properly)

I'll leave an approve for you to use at your discretion 👍

PR #543, opened 2024-02-23
Disable decals on player layers

toberge:


Fueredoriku (approved):
Good fix!

We still have the problem with the pan, but we'll fix that later

PR #549, opened 2024-02-23
(Re)enable shadows in training mode

toberge:


Fueredoriku (approved):
Can confirm it works 👍

PR #550, opened 2024-02-23
Disable augment placer in builds :(

toberge:
They don't work in built versions of the game, sadly. Too complicated to fix right now.


Fueredoriku (approved):
Seems to work!

PR #554, opened 2024-02-23
Fix ais to not use stair checks, add ai aggro when no real players

Fueredoriku:


toberge (approved):
woah this was a nice n quick fix desu

PR #558, opened 2024-02-24
Fix ai navigation issues, turn volume down a bit

Fueredoriku:


toberge (approved):
Seems like an improvement 👍

PR #559, opened 2024-02-25
Reduce mouse sensitivity when zoomed

toberge:
Was ridiculously hard to aim when zoomed in with a mouse. Reduced the look speed to **a tenth** of its previous value.


Fueredoriku (approved):
Yup, good change 👍

PR #562, opened 2024-02-29
Stylized shadows

Fueredoriku:
- Shadows now have a line texture to make them look more "hand-drawn"
- Players now have a shadow directly underneath them so player can gauge depth/xz-plane position better while in air
- Fixed first person weapons casting shadows

Before this PR:
- Close range:
- image

- Further away:
- image
 

With this PR:
- Close range:
- image

- Further away:
- image


ChrisKrane:
I love the hatching shadows, but do you know what might be causing these lines in some building shadows?
Image Sequence_019_0000

Fueredoriku:
> I love the hatching shadows, but do you know what might be causing these lines in some building shadows? 

Here is how the shadow looks in dev (Whis is wayyyy worse imo):
image

It is caused by unnconnected mesh geometry in the models that cast the shadows and can be prevented by either remodelling the house or *maybe* by baking global illumination in every scene. 

Basicly this new textured shadow actually mitigates some of the problems we see in dev. Those problems should also be resolved at some point, but they should be fixed in their own PR imo

ChrisKrane:
> It is caused by unnconnected mesh geometry in the models that cast the shadows and can be prevented by either remodelling the house or _maybe_ by baking global illumination in every scene (which takes actual days with my computer).
> 
> Basicly this new textured shadow actually mitigates some of the problems we see in dev. Those problems should also be resolved at some point, but they should be fixed in their own PR imo as that task is maybe unintuively disconnected from what I have done in this PR

Understood. Then I think it looks great! Really good work with the shadows 😄

ChrisKrane (approved):
> It is caused by unnconnected mesh geometry in the models that cast the shadows and can be prevented by either remodelling the house or _maybe_ by baking global illumination in every scene (which takes actual days with my computer).
> 
> Basicly this new textured shadow actually mitigates some of the problems we see in dev. Those problems should also be resolved at some point, but they should be fixed in their own PR imo as that task is maybe unintuively disconnected from what I have done in this PR

Understood. Then I think it looks great! Really good work with the shadows 😄

PR #564, opened 2024-03-02
Prevent AI players from obscuring current bids

toberge:
Offsets their target by 2 units


ChrisKrane (approved):
Good stuff. Seems like the ai can still obscure the current bid if they all stack on one platform, but they move out of the way soon enough, so i'm not gonna worry about that.

PR #565, opened 2024-03-02
Make loading screen progress fill whole chambers + more

toberge:
- filled half chambers before -_-
- use different secret name for the mower, isn't rhythmic no more
- add more loading screen tips
-


ChrisKrane (approved):
The loading icon looks so much better with whole chambers filled. Mower name and loading screen tips good :). Doesn't seem to be a problem with this PR, but where did the ammo for the seeker go in the tutorial map?

toberge:
> The loading icon looks so much better with whole chambers filled. Mower name and loading screen tips good :). Doesn't seem to be a problem with this PR, but where did the ammo for the seeker go in the tutorial map?

Those ammo boxes were removed temporarily since the gun reassembly doesn't work in builds.

PR #566, opened 2024-03-02
made sobel look better

SondreHus:
Made sobel look better.

Todo: The effect is a bit reliant on resolution at the moment, we might want to look into making the thickness of the lines and sensitivities of the effects more dynamic in response to the resolutioon of the screen. Pain......


toberge (approved):
LVGTM

PR #568, opened 2024-03-07
Input tool tip at start of every match

Fueredoriku:
- Make input tooltip/auction tutorial appear at start of every match
- Fix a typo in loadingscreen tip (Lazur is named "BFL" in game)


toberge (approved):
LGTM

PR #570, opened 2024-03-07
Fix correct winner being chosen (last robot standing)

Fueredoriku:


CraZyB1336 (approved):
Yeah, looks good.

Kinda silly that we still had this bug. 😝

PR #572, opened 2024-03-07
Add some support for ultra wide aspect ratios

Fueredoriku:
Made the different UIs scale bettwer with width


toberge (approved):
Seems to work quite well, though with some minor issues that we can fix later:
- this:  
  image
- and the tiling background in the loading screen should be larger but it's some minor pain to fix 😔

PR #575, opened 2024-03-12
Adjust balance of various gun parts

toberge:
- Lower the accuracy of the coil barrel (it would reliably hit on all 4 shots before, which is enough to down a player)
- Increase the damage of the lazur barrel to 35, making 3 hits enough for a kill, which should make it more powerful until we properly buff it


Fueredoriku (approved):
Looks good!
Seems the coil inaccuracies made the trails look better too imo 🤠👍

PR #576, opened 2024-03-12
Use audio groups for more sounds

toberge:
- audiogroupify lots of sounds
- add option for continuous pitch bending, which may sound better for some sounds
- add PlayDelayed to audiogroups
- and fix that annoying not-quite-zero-ammo-left bug


Fueredoriku (approved):
Seems to work like it should, code looks good 👍

Issue #138, opened 2023-03-29
Pause Screen


Fueredoriku:
A pause screen with access to
- Main Menu (Quit Match)
- Options

so in summary
- [x] #585
- [x] #586
- [x] #587
- [ ] #588

toberge:
Status: Stale branch with some code.
Unresolved feedback on said code: Add a physics state singleton so that other components can check whether the game is paused _without having to be told so_

PR #583, opened 2024-03-16
Add "pause" menu

### LINKED ISSUE #585: Pause menu (that doesn't pause the game) accessible from shooting rounds (at least)

### LINKED ISSUE #586: Option to exit to main menu from a match

toberge:
image

Barely solves #138 
Closes #586 
Closes #585

### DIFF
--- a/Assets/Prefabs/Input/Player.prefab
+++ b/Assets/Prefabs/Input/Player.prefab
@@ -95,6 +95,7 @@ MonoBehaviour:
   steppingIgnoreMask:
     serializedVersion: 2
     m_Bits: 1
+  ShouldNotRespondToInputs: 0
 --- !u!23 &71015680234268065
 MeshRenderer:
   m_ObjectHideFlags: 0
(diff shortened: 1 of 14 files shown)


Fueredoriku (changes requested):
Good feature! 😁👍
I have some feedback and one requested change:

Feedback:
- I think we should rename the "Not paused" title to just "Menu" or "Pause Menu"
- The keybind for the gamepadaction of opening the menu should probably be mapped to the right button and not the left. Whic would be "start[gamepad]" instead of "exit[gamepad]". Alternatively it could be both.

Requested change:
- Currently the pause menu breaks zooming. If zooming is performed and the player then pauses and continues the game, zooming will remain broken until player is respawned.

toberge:
Fixened

Fueredoriku (changes requested):
Some feedback:
- It seems you can't select between buttons with a gamepad
- Opening more than one menu will deselect all buttons for all players except the newest menu

And a bit more nickpicky:
- For gamepads specifically: When zoomed, if you open a menu, the hand will remain in zoomed position until you zoom in and out again (I suspect this is due to gamepads having toggle behaviour for zooming that is not accounted for) 
image

toberge:
Setting this as draft again since it's waaaay more work than I assumed 🙃

toberge:
Redone in #759

PR #589, opened 2024-03-17
Disable auto spawning playeridentities on network connection

Fueredoriku:


toberge (approved):
LGTM :)))))))

but why is my gamepad a network connection? because all input is treated as the same kinda thing?

PR #593, opened 2024-03-22
Diasble unfinished netcode to fix training

Fueredoriku:
Seems some of the unfinished etworj integration is stopping players from entering training grounds.
- The PR deactivates those components until the lobby system is actually implmented
- Adds some extra checks to the steamManager so it handles not having access to steam better
- Removes superflous eventsystem in training grounds


toberge (approved):
Finally it works again 🙃

PR #597, opened 2024-03-22
Fix and refactor crouch+jump transitions

### LINKED ISSUE #596: Fix crouch state machine

toberge:
Closes #596 

The bug turned out to be that the outer, frictionless collider _sometimes_ triggered the ground check at the apex of a leap jump. Fixed by ignoring the IgnoreRaycast layer in the ground check mask 💀


Assets/Scripts/Control&Input/PlayerMovement.cs
```diff
@@ -243,59 +277,68 @@ private void OnCrouch(InputAction.CallbackContext ctx)
...
         cameraCrouchPerformedTween = inputManager.gameObject.LeanMoveLocalY(localCameraHeight - crouchedHeightOffset, 0.2f).id;
         gunCrouchCanceledTween = gunHolder.LeanMoveLocalY(localGunHolderHeight - crouchedHeightGunOffset, 0.2f).id;
     }
 
-    private void EnableDash()
+    private void StopCrouching()
     {
-        StartCoroutine(JumpTimeout(leapTimeout));
+        isCrouching = false;
+        if (state is GroundState.Grounded)
```
Fueredoriku, line 303: I would like us not to do this if check. There is a little known feature I like a lot where if you cancel your crouch mid air then hold crouch again before you land to perform a high jump. (leaphop-cancelling basicly) I really enjoy the control this grants in some occasions (example below), but removing "leaphop-cancelling" is of course also an option if we decide we want to.

Here is an example where I removed the if check (leap jumping, letting go of crouch and then holding it again while in the same jump before perfoming a new jump):
CrouchJumpwithout
And one as is here (same inputs):
CrouchJumpwith

Fueredoriku (approved):
While the issue itself may have been small, it at least led to a very nice refactor 😁👍

I only have a small comment regarding an obscure control mechanic I think we should consider keeping, but other than that this seems like a great improvement, so I'll leave an approve for you to use at your discretion!

Great work!

Assets/Scripts/Control&Input/PlayerMovement.cs
toberge: Think I got the intention here so I merge this now >:)

PR #598, opened 2024-03-22
Make cannon barrel reasonably speedy

toberge:


Fueredoriku (approved):
I'm a bit afraid of it being too strong with speedboosts now, but I think we should demo this as is, might be just the right amount of crazy 🤠

PR #599, opened 2024-03-24
Add satelite uplink extension that calls down space garbage to crush your enemies

toberge:
image

https://github.com/hackerspace-ntnu/Red-Planet-Rampage/assets/44986729/3ecd0fd7-0d9d-49b0-89fe-e739f895b796

# All the stuff

Left out some polish that isn't absolutely necessary

- [ ] Models for basic assortment of garbage 
  - [x] Rock
  - [x] Anvil
  - [x] Computer
  - [x] Trash can (same as on the extension model)
  - [x] Barrel
  - [ ] Satelite parts?
- [x] Falling garbage
- [ ] Different stats for different stuff?
- [x] Shockwave??? (doesn't seem quite as necessary now that there's more stuff and it falls fast)
- [x] Sound effects (need editing and perhaps new recording)
- [x] Targeting reticle
- [x] Launch garbage on hit/ricochet
- [x] Only after cooldown
- [x] Extension model
- [ ] Texture extension model? (not really necessary, material works fine enough)
- ~~Some place to display wireframe on extension model???? on screen???~~
- [x] Fancier text??? le green hackur???
- ~~Wireframe shader~~
- [x] and of course uhhh limit the spam with a max cap
- [x] and perhaps pool the created garbage so it doesn't become ridonculous

# Synergies

Barrels have the following synergies with this trash uplink extension:
- Hatapult and lazur cause trash rain on every bounce/ricochet
- Cannon explosions cause orbital trash strikes on all hitboxes that were damaged by the explosion
- Shotgun and coil have more bullets per shot and thus cause heavy rain wherever they hit
- For clarity: Bullet barrel causes only one trash item to fall from the sky

# Misc changes

- All projectile controllers now create a new projectile state for each shot. Had issues with tracking fired projectiles, and this seemed to be the simplest way to make the tracking code work for all barrels.
- Removed the "not attached to gun parent" log messages since they clutter the console in auction and training mode, and this is expected behaviour at this point.
- Added webm version of the opening cinematic since the game doesn't run (in editor) on Linux with the previous format.
- Extracted radial timer code to a component so I didn't have to paste that stuff _another_ time smh


Assets/Scripts/Hazards/FallingHazard.cs
```diff
@@ -0,0 +1,63 @@
...
+        if (!isMovingFastEnoughToKill)
+            return;
+
+        isFirstImpact = false;
+
+        var instance = Instantiate(impactExplosion, transform.position, Quaternion.identity);
+        instance.Init();
+        instance.Explode(player);
+    }
+}
```
Fueredoriku, line 63: Missing endline

Assets/Scripts/Augment/BulletModifiers/KnockbackOnShotModifier.cs
```diff
@@ -17,7 +17,6 @@ private void Awake()
         gunController = transform.parent.GetComponent<GunController>();
         if (!gunController)
```
Fueredoriku, line 18: I would prefer we rewrote all these guncontroller checks to the abbreviated no-body version, like:
`
if (!gunController)
   return;
`
But I consider this a nitpick you don't need to do here unless you want to, as ideally later we won't need this pattern if we resolve the now infamous #126

Assets/Scripts/Augment/AugmentImplementations/SateliteUplink.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        var offset = Random.Range(30f, 0);
+        var launchPoint = target + (launchHeight + offset) * Vector3.up;
+
+        var garbage = spaceGarbage.RandomElement();
+        var garbageInstance = Instantiate(garbage, launchPoint, Quaternion.AngleAxis(Random.Range(0f, 360f), Vector3.up));
+        garbageInstance.Player = gunController.Player;
+
+        var targetInstance = Instantiate(targetingReticle, target, Quaternion.identity);
+        Destroy(targetInstance, 2); // TODO destroy only when we hit the ground (?)
```
Fueredoriku, line 140: Remove devcomment, and maybe consider objectpooling these? Could be a nice solution here

Fueredoriku (commented):
Very fun augment and many good refactors here 😁👍

I have a couple suggestions (and some nitpicks) in the comments, but I also have feedback regarding the augment I'll outlay here:

I think the amount of trash on the ground can become a bit too much over time and/or with certain combinations, both in terms of performance and gameplay. I think we should probably have an objectpool for the trash too (in addition to one for the targeting markers I commented about), and the pool should probably despawn trash after maybe 25 trash max(?) and/or if a certain amount of time has passed.

toberge:
Agreed, I'll create a trash pool (it was already in my checklist lol)

Assets/Scripts/Hazards/FallingHazard.cs
toberge: Thanks, had to turn off the LSP that provided autoformatting for performance reasons (vscode issues idk) so I missed that one :upside_down_face:

Fueredoriku (approved):
Very nice performance improvements, I think this is basicly ready to go now 😁👍

I only have a small nitpick, so I'll leave an approve for you to use at your discretion:
image
It seems the falling hazards are not put into a container parent object as the recticles are, which would be nice to not clog the scene hierarchy

Issue #577, opened 2024-03-12
Prevent Moiré effect on player head graphics


toberge:
The dots should be larger, we're displaying this stuff at a distance 😤

PR #600, opened 2024-03-26
Add mipmapping to player head icons

Fueredoriku:
Solves #577


toberge (approved):
Seems to fix the issue, but it comes with the caveat of not having a dotted look on lower resolutions (the player heads look like they're regularly filled)

Fueredoriku:
It's a good enough trade off I think, especially after testing some different alternatives (2*2, 3*3 and 4*4 checkerboards) that also produced the Moiré effect at noticable distances. Larger checkerboard patterns than that just looks bad in terms of pixel art imo

PR #601, opened 2024-03-26
Fix cinematic intro format so it works on Linux

toberge:
Add webm version of the opening cinematic since the game doesn't run (in editor) on Linux with the previous format.

Making the change here as well as in #599 since it could take a while to get that one merge-ready


Fueredoriku (approved):
Nice fix 🤠👍

PR #609, opened 2024-04-02
Add local stats for augment combinations

### LINKED ISSUE #612: Actually update combination stats

toberge:
Closes #612


toberge:
prevent users from quitting a match early and saving weird data

toberge:
> prevent users from quitting a match early and saving weird data

Won't do this now as it requires further work on the stat-storing code, which shouldn't be prioritized right now.

Assets/Scripts/Gamestate/Round.cs
```diff
@@ -119,7 +119,7 @@ private void CheckWinCondition(PlayerManager lastKiller)
     {
         if (livingPlayers.Count < 2)
         {
-            winner = livingPlayers.FirstOrDefault()?.identity;
+            winner = lastKiller?.identity;
```
Fueredoriku: Last living player is not necessarily last killer, is this change intended to change the logic or was it a refactor with unintended consequences?

Fueredoriku (changes requested):
Nice, refactors and changes 😁👍
 There is just one change I think may lead to unintended consequences which I think might need to be adressed before merge

Assets/Scripts/Gamestate/Round.cs
toberge: Was a refactor, the method's parameter was unused and this change technically resolved that. Reverted and removed the parameter.

Fueredoriku (approved):
Very nice, will be great for the improved future gallery 🤠

Issue #615, opened 2024-04-08
Syncing Players


Fueredoriku:
- Syncing inputs
- Syncing transforms
- Syncing animations

PR #616, opened 2024-04-08
Add working steamsockets connection and basic lobby UI

Fueredoriku:
Has a lot of jank, but the main point of this PR is the actual steamsockets connection, basic scene change for every client, and a UI from which one can host.

Also added a logger for the built game which we should disable at the end of milestone build.

Won't fix here, should be fixed in #615 instead:
- 1 additional player is spawned for host
- Multiple local players is not supported very well
- Splitscreen still applies to every player

Should be fixed other places:
- The join button currently does *nothing*, and won't be fixed before we make a better lobby UI.


toberge (approved):
I've got some concerns, but this needs further work anyways

PR #619, opened 2024-04-11
Feature/improve level select

Fueredoriku:


Assets/Scripts/UI/LevelCard.cs
```diff
@@ -0,0 +1,28 @@
...
+using UnityEngine;
+
+public class LevelCard : MonoBehaviour
+{
+    [SerializeField]
+    private GameObject levelCard;
+    [SerializeField]
+    private string cardName;
+
+    public GameObject getLevelCard()
```
toberge, line 12: These methods should be properties instead (e.g. LevelCard => ... for this one)

Assets/Scripts/UI/LevelSelectManager.cs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    private void SetupButtonNavigation(UnityEngine.UI.Button button, int index, int totalCount)
+    {
+        Navigation navigation = button.navigation;
+        navigation.mode = Navigation.Mode.Explicit;
+
+        if (index == 0 && totalCount > 1) //First levelcard button
+        {
+            navigation.selectOnRight = instantiatedCards[index + 1].GetComponent<UnityEngine.UI.Button>();
```
toberge, line 95: I'd prefer wraparound here (when on one end you end up at the other when you try to go "beyond"). We've got that in all other menus now I think? This also goes for the vertical movement to the back button. With this in mind, you should be able to use `(index +/- 1).Mod(totalCount)` as index in the `instantiatedCards` array, which also makes the if branches unnecessary.

Assets/Scripts/UI/MainMenu/MainMenuMoveCamera.cs
```diff
@@ -36,5 +43,20 @@ public void MoveToPlayerSelect()
...
+            LeanTween.sequence().append(LeanTween.rotateX(playerSelectCamera.gameObject, 9f, 1).setEaseInOutExpo());
+        }
+    }
+
+    public void MoveToLevelSelect()
+    {
+        inLevelSelect = true;
+        LeanTween.sequence().append(LeanTween.moveLocal(playerSelectCamera.gameObject, secondPosition.localPosition, 1).setEaseInOutExpo());
+        LeanTween.sequence().append(LeanTween.rotateX(playerSelectCamera.gameObject, 90f, 1).setEaseInOutExpo());
+        //68
```
toberge, line 60: This comment could be removed, I think?

toberge (approved):
Looks and feels very nice!

Only UI nitpick I've got is that I feel there should be a border or _something_ around the selected card, so that when you enter the level select menu it is more obvious that the card is the thing that is selected. And the cards could be a little closer together, perhaps, though that might just be me.

Assets/Scripts/UI/LevelCard.cs
ChrisKrane: I learn something new every PR. Fixed

Assets/Scripts/UI/LevelSelectManager.cs
ChrisKrane: Fixed wraparound

Assets/Scripts/UI/MainMenu/MainMenuMoveCamera.cs
ChrisKrane: Wopsie. fixed

PR #620, opened 2024-04-11
Bullet networking

toberge:
Transmit bullets through the wire

# Plan

- [ ] Simply make guns fire on all clients/servers when they are fired from one player
- [ ] Transform this into firing from point and in direction that is given by player
- [ ] Deal with allllll the hijinks (big one)
- [ ] Seed random operations (e.g. lightning bolts)

# ⚠️ Problems ⚠️ 

- NetworkIdentity must be parent of all NetworkBehaviours _when it awakens_ (this ain't the case for GunController)


toberge:
Replaced by #635

PR #625, opened 2024-04-26
Add procedural animated SDF clouds

Fueredoriku:
Now featuring:
- Prettier clouds
- More optimized (both in speed and size) and readable shader code

image


toberge (approved):
Good clouds 👍

PR #626, opened 2024-04-30
Make trash uplink deterministic (enough for netcode)

toberge:
- Avoid relying on unreliable physics for the fall of the space garbage
- Switch back to physics once the trash hits the ground
- Alter launch position horizontally to make the experience recognizably
  chaotic
- Should be possible to emit the point that was hit back to the modifier
  script for future network syncing

### DIFF
--- a/Assets/Scripts/Hazards/FallingHazard.cs
+++ b/Assets/Scripts/Hazards/FallingHazard.cs
@@ -1,11 +1,18 @@
+using System.Linq;
 using UnityEngine;
 
 public class FallingHazard : MonoBehaviour
 {
-    [SerializeField] private float deadlyVelocityThreshold = 100;
     [SerializeField] private float audibleVelocityThreshold = 50;
+    [SerializeField] private float initialFallVelocity = 10;
+
+    [SerializeField] private float spherecastRadius = 2;
+    [SerializeField] private LayerMask spherecastMask;
+
     [SerializeField] private AudioGroup soundEffect;
     [SerializeField] private ExplosionController impactExplosion;
+
+    private const float gravity = 9.81f; // >:)
 
     private PlayerManager player;
     public PlayerManager Player
@@ -17,9 +24,9 @@ public class FallingHazard : MonoBehaviour
     private Rigidbody body;
     private AudioSource audioSource;
 
-    private bool isFirstImpact = true;
-
-    private float lastVelocity = 0f;
+    private bool isFalling = false;
+    private float lastVelocity;
+    private float fallVelocity;
 
     private void Start()
     {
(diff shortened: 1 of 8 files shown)


Assets/Scripts/Hazards/FallingHazard.cs
```diff
@@ -44,11 +48,55 @@ public void Launch(Vector3 position)
...
+        if (!isFalling)
+            return;
+
+
+        // Accelerate
+        fallVelocity += gravity * Time.fixedDeltaTime;
+        var fallDistance = fallVelocity * Time.fixedDeltaTime;
+
+        // Check for collision in current distance
+        if (Physics.SphereCast(transform.position + Vector3.down, spherecastRadius, -transform.up, out RaycastHit hit, fallDistance, spherecastMask))
```
Fueredoriku: Since this happens every fixedupdate, I think it would be better to use overlap sphere in this case

Assets/Scripts/Hazards/FallingHazard.cs
```diff
@@ -1,9 +1,15 @@
+using System.Linq;
 using UnityEngine;
 
 public class FallingHazard : MonoBehaviour
 {
-    [SerializeField] private float deadlyVelocityThreshold = 100;
     [SerializeField] private float audibleVelocityThreshold = 50;
+    [SerializeField] private float initialFallVelocity = 10;
+    [SerializeField] private float gravity = 9.81f;
```
Fueredoriku: Not const and not mars gravity? 🤔
I actually think we might need to start using other gravities than mars', as I've recently read that other games tend to use way higher constants for earth gravity in games for better game feel, but I'll slot that discussion up for next meeting I think.
I think this probably should be const tho.

Fueredoriku (changes requested):
Seems like a good solution in terms of networking! 👍

The spherecast should be an overlapshere though, because the performance dies when you use a shotgun or a similar augment that spawns multiple fallinghazards.

I also noticed that with a shotgun, shooting directly below yourself will cause every shot except the first ones that hit you to teleport high into the air before falling down? Seemed like odd behaviour, but it could probably be left in with minimal impact, so don't prioritize fixing that

While I don't think you necessarily need to implement it in this PR, deterministic randomness with a seed would be great for the randomness here. We can of course implement that later, but I don't think that syncing hitpoints alone are enough in this case. This can of course wait for a later PR

Assets/Scripts/Hazards/FallingHazard.cs
toberge: Rewritten to use OverlapCapsule. Performance still takes a hit when launching trash, but should not be as significant as before. (didn't check the FPS counter when I tested the previous version)

Assets/Scripts/Hazards/FallingHazard.cs
toberge: Exposed it in order to tune it more easily, but that didn't turn out to be necessary as it already felt appropriate.
Made it a constant now.
Physically accurate Mars gravity is too weak for our purposes 😈

Fueredoriku (approved):
Nice!

After some testing with the profiler, I found the audio to be a suprisingly big culprit for the performance loss, specifically when it's called 11 times in quick succesion from onGround in fixedupdate with a shotgun. To further optimize, we could probably cap a maximum amount of audio that can be played from trash at the same time
image

I don't think we need to do that right now necessarily, as we could probably just put up a new issue instead. In the meantime I'll leave an approve for you to use at your discretion 🤠👍

toberge:
Since the audio issue isn't strictly related to this PR I'll merge it and add an issue for it. Seems like something we could improve in other places as well (including for the shotgun itself; it still plays 8x the same firing sound)

PR #628, opened 2024-05-02
Fix ContionousDamage destroying itself with optional parameter

### LINKED ISSUE #623: Fire does not despawn on target if owner dies

Fueredoriku:


toberge (approved):
LGTM

PR #629, opened 2024-05-06
Use seeded random for synced behaviour

toberge:
- [x] Add extension methods for System.Random
- [x] Switch to seeded System.Random in all components that should be synced (basically anything with a random position/rotation that needs to stay the same)


Fueredoriku (changes requested):
Very nice changes! 😎

The only problem I've found is that the random.UnitVector seems to have a weird offset? 🤔
Try shooting with the coil barrel and the shots seem to be aiming as if the forward vector was painting about 20 degress more upwards than it really is

Other than that I think all the other changes look good!

toberge:
> The only problem I've found is that the random.UnitVector seems to have a weird offset? 🤔 Try shooting with the coil barrel and the shots seem to be aiming as if the forward vector was painting about 20 degress more upwards than it really is

Thanks for finding the buggerino, I didn't test everything 🙃

Fueredoriku (approved):
Tested and seems to look good with newest changes🤠👍

PR #632, opened 2024-05-08
Tether polish

### LINKED ISSUE #624: Tether wire is invisible for other players than owner
- Should also be layered so it ignores moving/removable objects like exploding barrels

Fueredoriku:
- Fix wire layering #624 
- Add plug hitbox+health, you and other players can shoot at plug to unplug tether
- Make unplugging easier while in air
- Wire no longer collides with explosive barrels (as it looks very weird when they are removed)
- Extra tether coil animation for increasing or decreasing wire length
- "Plop"- soundFX when tether is unplugged


Assets/Prefabs/UI/LoadingScreen.prefab
```diff
@@ -146,6 +146,8 @@ MonoBehaviour:
   - If you can't outbid your opponents, the chips you don't spend may come in handy
     later.
   - Cowbots are, contrary to common belief, not cows.
+  - Both you and others can unplug the Tether augment by shooting at the plug, so
```
toberge: Perhaps "body" instead of "augment" here?

Assets/Scripts/Augment/AugmentImplementations/RopeBody.cs
```diff
@@ -56,16 +65,25 @@ public override void Start()
...
-            rope.enabled = false;
-            gunController.stats.Ammo = 0;
-            plugAnchor.gameObject.SetActive(false);
-        }
+        if (rope.RopeLength > ropeLength + 4f || (movement.StateIsAir && rope.RopeLength > ropeLength + 1f))
+            RemoveRope();
             
     }
 
+    private void KillRope(HealthController healthController, float damage, DamageInfo info)
```
toberge: Perhaps let it be an overload of RemoveRope instead, so the naming stays consistent.

toberge (approved):
Works really well now, and though I am sad to lose the ability to hang down from the platforms in Grand Canyon, the lowered breaking force feels much better than the previously unwieldy tugging (though perhaps it should be made a *little* stronger).

Only other functional feedback I've got is that the rope starts appearing *before* the plug has been placed when you reattach.

Assets/Prefabs/UI/LoadingScreen.prefab
Fueredoriku: Yup

Assets/Scripts/Augment/AugmentImplementations/RopeBody.cs
Fueredoriku: Fixed 👍

PR #635, opened 2024-05-09
Bullet networking (attempt 2)

### LINKED ISSUE #640: Synchronize that a gun should fire

toberge:
Replaces the old PR #620  cuz I don't wanna overwrite that branch yet

# Plan
- [x] Send info about other players to clients
- [x] Spawn and initialize players with guns
- [x] Simply make the correct gun fire on all clients/servers when it is fired by its owner
- [x] Use SteamID to identify players
- [x] Make the trail and so on actually show up on all clients

# Extra that should be done beyond this
- [ ] Transform this into firing from point and in direction that is given by player
- [ ] Deal with allllll the hijinks (big one)
- [ ] Provide seeds for random operations (e.g. lightning bolts)

Closes #640


Assets/Mirror/Core/NetworkIdentity.cs
```diff
@@ -294,10 +294,13 @@ internal static void ResetServerStatics()
...
         public bool SpawnedFromInstantiate { get; private set; }
 
-        // NetworkBehaviour components are initialized in Awake once.
-        // Changing them at runtime would get client & server out of sync.
-        // BUT internal so tests can add them after creating the NetworkIdentity
-        internal void InitializeNetworkBehaviours()
+        /// <summary>
+        /// Made public to support gun init. DO NOT USE IN OTHER PLACES, PLEASE.
+        /// NetworkBehaviour components are initialized in Awake once.
+        /// Changing them at runtime would get client & server out of sync.
```
Fueredoriku, line 300: Seems a bit sus, but it's an ok workaround. We might consider looking for other solutions in the future... maybe

Assets/Scripts/Augment/GunFactory.cs
```diff
@@ -53,19 +54,24 @@ public static GameObject InstantiateGun(Item bodyPrefab, Item barrelPrefab, Item
         if (displayGun.GunController.HasRecoil)
             firstPersonGunController.onFire += displayGun.GunController.PlayRecoil;
 
-        if (displayGun.GunController.projectile.GetType() == typeof(BulletController))
+        // Bypass the layer disabling here
+        // actually it probably isn't disabling, so keep it!
+        //if (isNetworkGun)
+            //return gun;
```
Fueredoriku: Maybe remove if not in use, or are you planning on iterating on this in a future PR?

Fueredoriku (approved):
Great stuff!! 🎉

I just have couple of comments, but nothing big, so I'll leave you an approve 👍

Great work!

Assets/Scripts/Augment/GunFactory.cs
toberge: Whoops, I forgor 💀

Assets/Mirror/Core/NetworkIdentity.cs
toberge: Dunno if we have that much choice here. We can't spawn in fully built guns through Mirror at least, as that would require several hundred different prefabs AFAICT. Leaving it like this for the time being.

Fueredoriku: But it's impossible to call the Setgun method for every client on each player after they are spawned on the client? 
If that doesn't work now I imagine it can work with a refactor maybe, where a networked guncontroller sits on the player on some transform on the player 🤔
I guess we can leave this as is for now, but we might need to look into it later

toberge: - Calling setgun after spawned on client wouldn't make any difference, as Awake is called on a NetworkIdentity when it is instantiated
- We want to network some modifiers as well, not just the guncontroller

Assets/Scripts/Augment/GunFactory.cs
toberge: Removed 🫡

PR #636, opened 2024-05-09
Add better looking background for lobbyUI, fix UI camera panning

Fueredoriku:
- Added better background for lobby choice ui menu
- Fix issue where start button wouldn't look selectable when only clients have joined lobby
- Fix issue where camera in player select would pan to level select when it shouldn't have

### DIFF
--- a/Assets/Scenes/Menu.unity
+++ b/Assets/Scenes/Menu.unity
@@ -38,7 +38,7 @@ RenderSettings:
   m_ReflectionIntensity: 1
   m_CustomReflection: {fileID: 0}
   m_Sun: {fileID: 785370674}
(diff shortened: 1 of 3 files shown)


toberge (changes requested):
Seems like a good change to separate the lobby choice from what you see while _in_ a lobby, but I've got a few UI nitpicks:

- The title should be spread out over one line (there's a lot of empty space to the sides)
- The buttons should be less tall in order to fit in with the rest of the menu (like 4/5ths to match the main menu, try making the buttons start at the same height and have a gap at the bottom maybe, or just space them lower)

image

Oh and AFAICT the buggerinos have been fixed 👍

Fueredoriku:
Fixed feedback, also added a camera pan for the lobby select to further differentiate it

toberge (approved):
Looks good now 👍

PR #644, opened 2024-05-14
Make host determine client players' health (and network local play)

### LINKED ISSUE #633: Network damage
- [ ] Players take damage as demanded by host
- [ ] Explosive barrels and soda cans explode/fly synchronized
- [ ] Other props take damage without syncing (e.g. the bendy cardboard cutouts, anything with the PushedOnHit component)

toberge:
Requires doing stuff to track players already.

# Stuff done:
- [x] Track which player ID corresponds to which PlayerManager in MatchController
- [x] Make HealthControllers on server emit damage events to client players
- [x] Make explosive barrels and soda cans networked as well
  - This broke local play and forced me to make this PR humongous in order to avoid that issue
- [x] Synchronize visual aim angle
- [x] Synchronize actual shooting angle (perhaps avoid using a bloody transform)
- [x] Make bidding and match progress work in networked mode _locally on the host machine_ for now
  - This was needed in order to avoid breaking local play
- [x] Fix whatever it is that prevents joining with local player while in the lobby (works when joining _before_ you enter lobby)
- [x] Make AI players work locally (current jank: dunno)
- [x] Transfer loadout and chip info between rounds (oh joy)
- [x] Ensure that ChangeInputMaps actually runs when changing scene on clients
- [ ] Refactor player init
- [ ] Make AI players work online :))))))
- [x] Make scoreboards work online???
- [x] Use IDs in Round objects so that statistics make sense throughout the match? (IMPORTANT now)
- [ ] Make effects work, perhaps?
  - This should be saved for later :/

# Problems encountered:
- [x] NetworkIdentities crash the game when you're not networked. Woah. Looking into fixing this. (fixed by making everything networked)
- [x] Everything needs to be networked (enough to not break local play)
- [x] Something causes things to go absolutely bonkers after third bidding round
  image
- [x] TRAINING MODE DOES NOT WORK :)))))
- ~~Shotgun got extremely powerful somehow, needs fixing~~ no, it didn't (probably)
- AI players are a conundrum
- MatchController and the various ways we set up players is a little hard to reason about

Closes #633


Assets/Mirror/Core/NetworkIdentity.cs
```diff
@@ -338,6 +338,15 @@ void ValidateComponents()
...
         {
+            // THIS IS A PATCH
+            // Bugger off if we ain't networked
+            //if (!NetworkManager.singleton.isNetworkActive)
+            //{
+            //    // TODO this ain't enough
+            //    Destroy(this);
+            //    return;
+            //}
+
```
Fueredoriku: Maybe remove this?

Assets/Scripts/Control&Input/Peer2PeerTransport.cs
```diff
@@ -1,32 +1,39 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
```
Fueredoriku: Is the plan to use regex to parse inventory strings for clients?

Assets/Scripts/Control&Input/Peer2PeerTransport.cs
```diff
@@ -65,175 +98,364 @@ public InitializeFPSPlayerMessage(uint id, Vector3 position, Quaternion rotation
...
+    {
+        identity.id = playerDetails.id;
+        var playerName = playerDetails.name;
+        if (players.Values.Where(p => p.steamID == playerDetails.steamID).Count() > 1)
+            playerName = $"{playerName} {playerDetails.localInputID + 1}";
+
+        identity.playerName = playerName;
+        identity.color = playerDetails.color;
+
+        // TODO avoid update invocation?
```
Fueredoriku: Might not need to do so given that we have a loading screen that covers over things like the chip popup soon while waiting for all clients. (I can't think of any other consequences right now at least)

Assets/Scripts/Control&Input/Peer2PeerTransport.cs
```diff
@@ -327,14 +552,108 @@ private IEnumerator WaitAndInitializeFPSPlayer(InitializeFPSPlayerMessage messag
...
+
+            // The identity sits on the input in this case, so edit that
+            var identity = input.GetComponent<PlayerIdentity>();
+            UpdateIdentityFromDetails(identity, playerDetails);
+        }
+        else if (playerDetails.type is PlayerType.AI && NetworkServer.active)
+        {
+            Debug.Log($"Spawning AI player {playerDetails.id}");
+            AIManager manager = player.GetComponent<AIManager>();
+            manager.SetLayer(3);
```
Fueredoriku: I've spotted multiple manager.SetLayer(3) in this file, maybe make a const named after the layer?

Assets/Scripts/Control&Input/PlayerInputManagerController.cs
```diff
@@ -66,12 +72,26 @@ public void RemoveListeners()
 
     private void OnPlayerJoined(PlayerInput playerInput)
     {
+        // TODO refactor this for online (should not source info from here)
```
Fueredoriku, line 75: If we always technically always have an active server now, we could consider this OK I think

Assets/Scripts/Gamestate/MatchController.cs
```diff
@@ -80,10 +62,11 @@ public class MatchController : MonoBehaviour
 
     private string currentMapName;
 
-    private List<Player> players = new List<Player>();
-    public List<Player> Players => players;
-    public IEnumerable<Player> AIPlayers => players.Where(p => p.playerManager is AIManager);
-    public IEnumerable<Player> HumanPlayers => players.Where(p => p.playerManager is not AIManager);
+    public Dictionary<uint, PlayerManager> PlayerById = new();
+    private List<PlayerManager> players = new();
+    public List<PlayerManager> Players => players;
```
Fueredoriku: Maybe we should use the ReadOnlyList for the public "Players" property in specific?

Assets/Scripts/Gamestate/PlayerManager.cs
```diff
@@ -135,14 +136,38 @@ private void Start()
...
             GunHolder.transform.forward = inputManager.transform.forward;
+            UpdateAimAngleCmd(movement.AimAngle.y);
+        }
+    }
+
+    // Hijinks for syncing aim from server to players
+    [Command]
+    private void UpdateAimAngleCmd(float yAngle)
+    {
+        UpdateAimAngleRpc(yAngle);
```
Fueredoriku, line 158: A downside of this syncing is that it sometimes seem to lead to behavior like: make players at their own hitbox if the aim down while falling. At least I think this might be the culprit, or is this a method for syncing all other players except the local client? 🤔

If this is also used to sync the local player aim, we might have to consider to add some kind of extra check or similar to avoid that aiming consequence

Assets/Scripts/Utils/SteamManager.cs
```diff
@@ -229,7 +218,8 @@ public void LeaveLobby()
 
     public bool ChangeScene(string sceneName)
     {
-        if (!isSteamInitialized || !transportProtocol.isNetworkActive)
+        // TODO avoid binding this to steam
+        if (!transportProtocol.isNetworkActive)
```
Fueredoriku: Since ChangeScene only needs access to trasnportProtocol, also known as the Peer2PeerTransport.singleton, we might as well already move this method straight to the Peer2PeerTransport class I think 👍

Assets/Scripts/HealthController.cs
```diff
@@ -33,7 +34,28 @@ private void Start()
 
     public void DealDamage(DamageInfo info)
     {
+        if (!isNetworked || currentHealth <= 0)
```
Fueredoriku: Why `currentHealth <= 0`? 🤔

Fueredoriku (commented):
Great (and huge) feature! 🤠
I very much agree with the way health controllers are synced at the moment (server authorative) and the consequences like local server even when not online (might prove to be an upside as we are more likely find unintended network behavior when testing normaly). These comments are mostly tied to stuff I found in the code.

I also have found these issues which, ranked in severity from highest first to lowest last:

- The guncontroller no longer works in Training grounds 

- Seems like level select manager no longer properly updates names when hosting a lobby, you get player 1 instead of steam name when on lobby start.

- Seems like the host client doesn't send the correct PlayerDetailsMessage, given that other client recieve the host as the default color, I'm guessing the host sends the playerIdentity that is on the host's player object and not the playerIdentity that is on the playerInput.

Assets/Scripts/Control&Input/Peer2PeerTransport.cs
toberge: Nope, that's vscode being overly helpful and bringing in a dependency for something I didn't wanna use 😒

Assets/Scripts/Control&Input/Peer2PeerTransport.cs
toberge: Useless comment since I avoided using the updatechip method, I think

Assets/Scripts/Utils/SteamManager.cs
toberge: Yeah, I was about to do that after fixing this condition, but didn't get around to refactoring all the scene change calls yet.

Assets/Scripts/HealthController.cs
toberge: Just thought it wouldn't make sense to send messages about hits to objects with negative/no health

Assets/Scripts/Control&Input/Peer2PeerTransport.cs
Fueredoriku: Aight, then I have no qualms with the planned implementation 🤠

Assets/Scripts/Control&Input/Peer2PeerTransport.cs
Fueredoriku: Nice, just resolve this after deletion in that case 👍

Assets/Scripts/HealthController.cs
Fueredoriku: I guess the practicality of this is that when you deal damage to something that is already dead, it won't necessarily be synchronized between clients. After thinking a bit about it, I can't come to think of any situation where that would lead to a problem, so this is probably fine 👍

Assets/Scripts/Control&Input/PlayerInputManagerController.cs
toberge: This will be overwritten/unused anyways, so I would like to eliminate this down the line

Assets/Scripts/HealthController.cs
toberge: It would lead to issues if we wanted to sync things like cacti and spent soda cans...
For exploding soda cans we might need to do this already, actually, since they die twice in order to explode 😔

toberge:
I also notice that we absolutely need to disable the Host button when not connected to Steam, as it absolutely doesn't work when not 😅

Assets/Scripts/Control&Input/PlayerInputManagerController.cs
Fueredoriku: Nice

Assets/Scripts/HealthController.cs
toberge: Removed this check for the reason state above. Soda cans should work now (though their position should be synced)

Assets/Scripts/Utils/SteamManager.cs
toberge: Only got a couple usages of this method left now

Assets/Scripts/Gamestate/PlayerManager.cs
toberge: The `if (inputManager)` guard in the RPC _should_ prevent this update from happening on the client where the input happened. Tested right now, and it _does_ seem to guard against that. The issue is likely to have a different cause then 🙃

Fueredoriku: Rubberbanding is probably the cause then I would guess

Assets/Scripts/Gamestate/MatchController.cs
toberge: Done. Should be done in lots of other places, but I'm saving that for later >:)

Assets/Scripts/Utils/SteamManager.cs
toberge: THE TARGET HAS BEEN ELIMINATED 🫡

Assets/Scripts/Utils/Scenes.cs
```diff
@@ -0,0 +1,7 @@
+public class Scenes
+{
+    public const string Menu = "Menu";
+    public const string ClientLobby = "ClientLobby";
+    public const string TrainingMode = "TrainingMode";
+    public const string Bidding = "Bidding";
+}
```
Fueredoriku: Add endline 🤠

Fueredoriku (approved):
Very nice, everything seems to work great! 🥳
Now we only need to fix synced bidding (including syncing player colors in bidding which seems to work at host player and not clients atm)

We are really close to wrapping this up now, this is great!

Assets/Scripts/Gamestate/PlayerManager.cs
toberge: Looking into this later ™️

PR #646, opened 2024-05-16
Add animation to secret names in weapon select

Fueredoriku:
Animation for secret names in weapon select, excluding "The Starter".

secretName


toberge (approved):
The animation looks fancy-scmancy enough 👍 
Only issue is that some of the hues are a little too dark (blue appears darker than other colors even though the saturation/lightness is the samme), but this ain't that important.
With this new fancy text, it becomes even more apparent that we should center the gun name vertically inside the name box, since it looks a little awkward right now when the names covers only one line.

Fueredoriku:
Added #652 so we remember to fix this later (tm)

PR #651, opened 2024-05-16
Add toggleable crouch, add extra gamepad buttons for zoom+crouch

Fueredoriku:
- Add proper toggleable zoom and crouch alternatives
- Left gamepad trigger is now "hold to zoom" instead of "hold to crouch" (as per feedback and agreed upon earlier)
- Gamepad "hold to crouch" is now the previously unbound left gamepad shoulder button (the small one above the trigger)
- East gamepad button and left gamepad stick "push in" is now also toggleable crouch

### DIFF
--- a/Assets/ScriptableObjects/Input/PlayerControls.inputactions
+++ b/Assets/ScriptableObjects/Input/PlayerControls.inputactions
@@ -540,9 +540,27 @@
                     "initialStateCheck": false
                 },
                 {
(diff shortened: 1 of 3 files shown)


toberge (approved):
Seems good (though I'm a little sad that we lose crouch on left trigger, I know people have been requesting zoom to be bound there for ages :)))))))

PR #653, opened 2024-05-19
Fix #584, fix color correction for diffuse, add better shadows

### LINKED ISSUE #584: Remove specular in shadows

Fueredoriku:
Also fixes a problem where color of texture would impact diffuse lighting


toberge (approved):
We still have some ways to go with these kinds of edgecases, but things look so much better now 🎉 
image
image
image

PR #654, opened 2024-05-19
Add client lobby for non-host players in steam lobbies

### LINKED ISSUE #642: Joined online lobby scene
Add lobby scene for clients that joined a lobby hosted by another client.
- Ability to see other clients 
- Ability for a client to quit a joined lobby

Fueredoriku:


toberge (changes requested):
Before merging this, I think we should put the poker room in a prefab along with the Players and PlayerNames objects, as these are required for the lobby room and should be kept the same(ish) between the main menu and the client lobby scene. The terrain and so on from the main menu should be removed from this scene as well. Illustrated in the following screenshot:

image

toberge (approved):
Seems good now! I'm just a wee bit puzzled by ye ol' Position1 and Position2 right here: (I see one of them is referenced in the lobby script)

image

PR #658, opened 2024-05-21
Add join menu for all available match servers

Fueredoriku:


Assets/Scripts/Utils/SteamManager.cs
```diff
@@ -37,13 +38,20 @@ public class SteamManager : MonoBehaviour
     public List<ulong> PlayerIDs = new();
     public delegate void LobbyEvent();
     public LobbyEvent LobbyPlayerUpdate;
+    public LobbyEvent LobbyListUpdate;
+    private List<CSteamID> lobbies = new();
+    public Dictionary<CSteamID, string> Lobbies = new();
```
toberge, line 43: Ideally the `string` value should be an object with s'more interesting values (and the public access could be the .Values of the dict if the key was part of that dict, and the dict key could be the ulong value that is encapsulated in the CSteamID), but I'll let that slide for now as this menu needs a refactor for player count and so on later.

toberge (approved):
Seems to work, though I managed to get a curious result once, where my own lobby showed up along with an unnamed one, when I had started a lobby on a dummy account:

image

The comment I've left is merely a suggested improvement that can be done once we need more data about these lobbies

Assets/Scripts/Utils/SteamManager.cs
Fueredoriku: Yeah, should be done as soon as we want any other data that is not just a string name

PR #660, opened 2024-05-25
Synchronize auction

toberge:
# in this PR we synchronize
- [x] Weapon construction readiness
- [x] Items assigned to bidding platforms
- [x] Placing bids
- [x] Receiving items
- [x] Loading screen

# in this PR we leave these problems
- The red cross shows up on the wrong bidding sign when a remote player can't bid (idk how)

### DIFF
--- a/Assets/Scripts/Control&Input/ItemSelectManager.cs
+++ b/Assets/Scripts/Control&Input/ItemSelectManager.cs
@@ -1,18 +1,36 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using Mirror;
 using UnityEngine;
 
-public class ItemSelectManager : MonoBehaviour
+// TODO remove first two structs if unused
+internal struct StartTrackingMessage : NetworkMessage
+{
+}
+
+internal struct ReadyToTrackMessage : NetworkMessage
+{
+}
+
+internal struct ClientReadyMessage : NetworkMessage
+{
+}
+
+internal struct ClientNotReadyMessage : NetworkMessage
+{
+}
+
+public class ItemSelectManager : NetworkBehaviour
 {
     [SerializeField] private float graceTime = 1;
 
     private List<ItemSelectMenu> itemSelectMenus;
+    private Dictionary<int, bool> clientReadyByID = new();
 
     private Coroutine waitRoutine;
 
-    [SerializeField]
-    private GameObject loadingScreen;
+    [SerializeField] private GameObject loadingScreen;
 
     private int delayDuration = 2;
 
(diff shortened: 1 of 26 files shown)


Assets/Scripts/Control&Input/ItemSelectManager.cs
```diff
@@ -1,19 +1,35 @@
...
-public class ItemSelectManager : MonoBehaviour
+// TODO remove first two structs if unused
+internal struct StartTrackingMessage : NetworkMessage
+{
+}
+
+internal struct ReadyToTrackMessage : NetworkMessage
+{
+}
+
```
Fueredoriku, line 15: Maybe remove the two above unless they are important for #661 ?

Assets/Scripts/Control&Input/Peer2PeerTransport.cs
```diff
@@ -325,24 +392,26 @@ public override void OnServerChangeScene(string newSceneName)
...
+                // From auction, so no need to update *here*
                 NetworkServer.ReplaceHandler<SpawnPlayerMessage>(OnSpawnFPSPlayer);
                 break;
         }
 
         base.OnServerChangeScene(newSceneName);
     }
 
+    // TODO move to ItemSelectManager (?)
     private void UpdateLoadout()
```
Fueredoriku, line 415: I assume this might be tweaked in #661, so I'll ignore this if that's the case

Assets/Scripts/Gamestate/PlayerManager.cs
```diff
@@ -253,14 +254,16 @@ public void SetPlayerInput(InputManager playerInput)
...
 
-    public void ApplyColorFromIdentity()
+    public void ApplyIdentity()
     {
         // Set player color
         meshBase.GetComponentInChildren<SkinnedMeshRenderer>().material.color = identity.color;
+        if (playerIK.TryGetComponent<BiddingPlayer>(out var biddingPlayer))
+        {
+            biddingPlayer.SetIdentity();
+        }
```
Fueredoriku, line 266: Seems a bit weird to place this here, I suspect there might be a cleaner way to do this if we refactor playerfactory (probably a lot, so maybe ignore for this PR)

Fueredoriku (approved):
Only have some small feedback here, as we have already moved on to #661

Assets/Scripts/Control&Input/Peer2PeerTransport.cs
toberge: I intend to do this thing, likely in #661

Assets/Scripts/Gamestate/PlayerManager.cs
toberge: Likely not that weird, as we would like to ensure that we listen to certain delegates from BiddingPlayer _only_ on the correct PlayerIdentity.

Assets/Scripts/Control&Input/ItemSelectManager.cs
toberge: Removing in #661 as these are not needed anymore

PR #661, opened 2024-05-25
Synchronize augments

toberge:
kinda depends on auction sync, but not

# this PR synchronizes
- [x] Portal extension (pos+rot of portal)
- [x] Trash uplink (explosion points) 
  - works quite well now
  - ⚠️ not sure if this is necessary, only visual and should be similar
- ~~Position + direction of shots explicitly~~ things work well w/o this!
  - as we can't rely on having the output be positioned and oriented the same on host and client
  - ⚠️ may require some extra safeguards in the future ™️ 
  - ⚠️ requires somewhat major refactor
- [x] Rubber sniper (possibly fixed by the above)
# this PR seeds randomness to
- [x] Zigzag modifier (coil)
- [x] Trash uplink
- [x] Bullet projectiles (for angle)
# this PR does not fix
- shotgun madness on explosive barrel (check ondeath)
- synced unplugging of tether
- Flying platforms
  - ⚠️ this seems way harder than I thought :o

### DIFF
--- a/Assets/Scripts/Augment/GunController.cs
+++ b/Assets/Scripts/Augment/GunController.cs
@@ -172,10 +172,6 @@ public class GunController : NetworkBehaviour
     private void RpcFire(Quaternion rotation)
     {
         onFireStart?.Invoke(stats);
-        // TODO does this work? if so, remove output
-        //if (authority)
-        //projectile.projectileOutput = outputs[0];
-        //projectile.projectileOutput = outputs[0];
         projectile.projectileOutput = outputs[0];
         projectile.projectileRotation = rotation;
         ActuallyFire();
--- a/Assets/Scripts/Augment/GunController.cs
+++ b/Assets/Scripts/Augment/GunController.cs
@@ -172,6 +172,10 @@ public class GunController : NetworkBehaviour
     private void RpcFire(Quaternion rotation)
     {
         onFireStart?.Invoke(stats);
+        // TODO does this work? if so, remove output
+        //if (authority)
+        //projectile.projectileOutput = outputs[0];
+        //projectile.projectileOutput = outputs[0];
         projectile.projectileOutput = outputs[0];
         projectile.projectileRotation = rotation;
         ActuallyFire();
--- a/Assets/Scripts/Control&Input/Peer2PeerTransport.cs
+++ b/Assets/Scripts/Control&Input/Peer2PeerTransport.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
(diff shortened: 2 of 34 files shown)


Assets/Scripts/Augment/GunController.cs
```diff
@@ -172,6 +172,10 @@ private void CmdFire(Quaternion rotation)
     private void RpcFire(Quaternion rotation)
     {
         onFireStart?.Invoke(stats);
+        // TODO does this work? if so, remove output
+        //if (authority)
+        //projectile.projectileOutput = outputs[0];
+        //projectile.projectileOutput = outputs[0];
```
Fueredoriku: I guess it does work? Maybe remove?🤔

Assets/Scripts/Control&Input/Peer2PeerTransport.cs
```diff
@@ -131,33 +131,43 @@ public class Peer2PeerTransport : NetworkManager
...
-        base.OnStartServer();
-        NetworkServer.RegisterHandler<PlayerConnectedMessage>(OnSpawnPlayerInput);
-        NetworkServer.RegisterHandler<UpdateLoadoutMessage>(OnReceiveUpdateLoadout);
+    public delegate void ConnectionEvent(int connectionID);
+    public ConnectionEvent OnDisconnect;
 
-        // Reinitialize player lookups
+    private void ResetState()
+    {
+        Debug.Log("Reset state");
```
Fueredoriku: I think we can remove this

Assets/Scripts/Extensions/PortalExtension/PortalExtensionController.cs
```diff
@@ -15,7 +12,7 @@ public class PortalExtensionController : GunExtension
 
     private Transform portalMover;
 
-    private GunController controller;
+    private GunController gunGontroller;
```
Fueredoriku: gunController instead? 😅

Fueredoriku (approved):
Seems to work, ver nice work! 😎
Only thing I want to mention is that we no longer have "player 1" "player 2" etc, and just write "player" for everyone when playing localy. I think we should at least set up a new issue for this

Assets/Scripts/Extensions/PortalExtension/PortalExtensionController.cs
toberge: Oh no, the gun gon bad

Assets/Scripts/Control&Input/Peer2PeerTransport.cs
toberge: Agreed 😅

Assets/Scripts/Augment/GunController.cs
toberge: Whoops, that's an attempt I forgot about

toberge:
> Seems to work, ver nice work! 😎 Only thing I want to mention is that we no longer have "player 1" "player 2" etc, and just write "player" for everyone when playing localy. I think we should at least set up a new issue for this

Regarding the player x problem, I'll likely deal with it in the #665 PR since I already started fixing problems with player select there

PR #665, opened 2024-05-27
Make it possible to leave lobbies ++

### LINKED ISSUE #662: Leaving/disconnecting during match
- Leaving lobby should stop the lobby and the network client/server, and clear all data in NetworkManager
- Disconnecting should do the same
- Having another player disconnect should leave their player data but remove their connection data, so that the game continues working and they may reconnect again (but we don't wait for them) (more advanced though) (see #672)

### LINKED ISSUE #670: Finishing a match causes it to be impossible to reverse the host-client relationship
- player A hosts
- player B joins
- they play a match
- player B hosts
- player A joins
- player A disconnects after a little while, and never shows up visually in the lobby

### LINKED ISSUE #671: Finishing a p2p match causes it to be impossible to see other lobbies to join
The lobby list is just always empty after you've been through a match

toberge:
Depends on #661 

Closes #670 
Closes #671 
Closes #662 

# Stuff done
- [x] Leave lobby when you leave the lobby
- [x] Leave lobby when you disconnect
- [x] Leave lobby when match is over (ideally you'd be back in the lobby again ready for a new match)
- [x] Players visually leave/enter lobbies (PlayerSelectManager jank fixed/worked around)
- [x] Clean up lobby leaving code so you rather have OurNetworkManagerThatIsYetToBeRenamed -> SteamManager
- [x] Disconnects during matches don't break the game (presently the dummy left by the escapee is unkillable, likely due to authority issues)
- [x] Show correct names (Player 1, 2, etc) in lobby
- [x] Fix refresh of lobby list after lobby exit or match end
- [x] Show loading screen when entering training mode
- [x] Fix chamber-filling animation on loading screen (it only played once 💀)
- [x] Fix spawnpoint selection (it used player ID, which can be > 3 now)

# Stuff planned but not done in this PR
- Use SetLobbyJoinable(false) to prevent users from joining an ongoing match and so on

### DIFF
--- a/Assets/Scripts/Control&Input/Peer2PeerTransport.cs
+++ b/Assets/Scripts/Control&Input/Peer2PeerTransport.cs
@@ -3,7 +3,6 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
 using Mirror;
-using OperatorExtensions;
 using Unity.VisualScripting;
 using UnityEngine;
 using UnityEngine.SceneManagement;
@@ -651,12 +650,18 @@ public class Peer2PeerTransport : NetworkManager
         StartCoroutine(WaitAndInitializeFPSPlayer(message));
     }
 
-    private void UpdateIdentityFromDetails(PlayerIdentity identity, PlayerDetails playerDetails)
+    // TODO move this somewhere else?
+    public static string PlayerNameWithIndex(PlayerDetails playerDetails)
     {
         var playerName = playerDetails.name;
         if (players.Values.Count(p => p.steamID == playerDetails.steamID) > 1)
             playerName = $"{playerName} {playerDetails.localInputID + 1}";
-        identity.UpdateFromDetails(playerDetails, playerName);
+        return playerName;
+    }
+
+    private void UpdateIdentityFromDetails(PlayerIdentity identity, PlayerDetails playerDetails)
+    {
+        identity.UpdateFromDetails(playerDetails, PlayerNameWithIndex(playerDetails));
     }
 
     private IEnumerator WaitAndInitializeFPSPlayer(InitializePlayerMessage message)
--- a/Assets/Scripts/Control&Input/Peer2PeerTransport.cs
+++ b/Assets/Scripts/Control&Input/Peer2PeerTransport.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
 using Mirror;
+using OperatorExtensions;
 using Unity.VisualScripting;
 using UnityEngine;
 using UnityEngine.SceneManagement;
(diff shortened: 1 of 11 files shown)


Assets/Scripts/Control&Input/Peer2PeerTransport.cs
```diff
@@ -216,6 +251,23 @@ public override void OnClientDisconnect()
             ResetState();
     }
 
+    private void OnPlayerLeft(PlayerLeftMessage message)
+    {
+        if (!players.TryGetValue(message.id, out var playerDetails))
+        {
+            Debug.LogError($"Received leave message for invalid player {message.id}");
```
Fueredoriku, line 259: We should probably consider logging some of these to players instead of just debug.log, we can probably set up a separate issue for that though

Assets/Scripts/Control&Input/Peer2PeerTransport.cs
```diff
@@ -254,10 +306,17 @@ public static void StartTrainingMode()
...
         while (!NetworkClient.isConnected && !singleton.isNetworkActive && players.Count < PlayerInputManagerController.Singleton.LocalPlayerInputs.Count)
             yield return new WaitForEndOfFrame();
         singleton.ServerChangeScene(Scenes.TrainingMode);
-        // TODO until after spawned
+
+        yield return new WaitForSeconds(LoadingScreen.Singleton.MandatoryDuration);
+
+        // Wait for player(s) to have spawned
+        while (FindObjectsByType<PlayerManager>(FindObjectsSortMode.None).Count() < players.Count)
+            yield return new WaitForEndOfFrame();
```
Fueredoriku, line 321: Hmmm, is this safe if something wrong happens during initialization? Maybe we need to put in a timeout after x seconds? Probably not too bad if it's only for trainingmode, but still

Assets/Scripts/Control&Input/Peer2PeerTransport.cs
```diff
@@ -318,6 +379,10 @@ private void OnSpawnPlayerInput(NetworkConnectionToClient connection, PlayerConn
             connection.Send(new InitialPlayerDetailsMessage(existingPlayer));
         }
 
+        // Pick among the available colors
+        if (!availableColors.TryPop(out var color))
+            color = Color.white;
```
Fueredoriku: Maybe just reset the color stack the same way as in the ResetState() instead of having white as backup?
Duplicate colors are a better backup imo

Fueredoriku (approved):
Seems good! I have a few comments, but I'll leave an approve for you to use at your discretion 😁👍

Assets/Scripts/Control&Input/Peer2PeerTransport.cs
toberge: Yeah, don't we already have an issue for killfeed?

Assets/Scripts/Control&Input/Peer2PeerTransport.cs
toberge: We do need a timeout for this and for the other waiting-for-spawn loops in this class and MatchController. The game *will* just appear to not load if there was an error while spawning players that caused less of them to be spawned.
I just consider this to be less of a problem than the other things that this PR fixes. See #675 (created right now)

PR #668, opened 2024-05-28
Fix training mode weapon switching (WIP)

toberge:


toberge:
Already done otherwise.

PR #669, opened 2024-05-28
Fix hat barrel build issue

### LINKED ISSUE #666: Firing the hat barrel may cause disconnect-level crash in builds
Stacktrace from a match against AIs where one AI had the hat barrel:

```
Disconnecting connection: connection(0) because handling a message of type Mirror.RpcMessage caused an Exception. This can happen if the other side accidentally (or an attacker intentionally) sent invalid data. Reason: System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
  at System.Collections.Generic.List`1[T].get_Item (System.Int32 index) [0x00009] in <cbc72d4a9767498db39486e941a498e3>:0 
  at HatBarrelModel.OnFire (GunStats stats) [0x0003e] in <912f8d6f363c4562a12a2363c79374ca>:0 
  at MeshProjectileController.InitializeProjectile (GunStats stats) [0x00023] in <912f8d6f363c4562a12a2363c79374ca>:0 
  at GunController.ActuallyFire () [0x0000c] in <912f8d6f363c4562a12a2363c79374ca>:0 
  at GunController.UserCode_RpcFire__Quaternion (UnityEngine.Quaternion rotation) [0x00036] in <912f8d6f363c4562a12a2363c79374ca>:0 
(...)
```

toberge:
Closes #666 (see stack trace for why I think this is the solution)

Note that it isn't possible to build the dev branch at the moment

### DIFF
--- a/Assets/Scripts/Augment/AugmentImplementations/Hat/HatBarrelModel.cs
+++ b/Assets/Scripts/Augment/AugmentImplementations/Hat/HatBarrelModel.cs
@@ -54,9 +54,9 @@ public class HatBarrelModel : AugmentAnimator
             isLastShot = false;
             return;
         }
-            
+
         isLastShot = stats.Ammo - 2 == 0;
-        for (int i = stats.Ammo - 2; i < magazineSize; i++)
+        for (int i = Mathf.Max(0, stats.Ammo - 2); i < magazineSize; i++)
         {
             ammunition[i].SetActive(false);
         }


Fueredoriku (approved):
Good find! 👍

toberge:
I suspect there's some more jank to find here (like, the reason why we had to subtract 2 in the first place)

PR #676, opened 2024-05-31
Enable 3D sound 🎉 

### LINKED ISSUE #649: 3D sound for non-split screen
Because of the restriction of only having a single audio listener if we are to use 3D noise, we have to toggle every audio source to 3D when NOT in splitscreen (or the opposite ofc)

toberge:
Current issues:
- Lazur barrel plays its firing sound twice when holding the trigger (on semi/full auto)


Assets/LazurController.cs
```diff
@@ -30,15 +32,26 @@ protected override void Awake()
...
+    private void Start()
     {
         Vfx.SetGraphicsBuffer("StartEndPositions", encoder.StartEndPositionsBuffer);
     }
 
-    public void FireLazur()
+    private void FireLazur()
     {
+        // TODO remove :)
+        Debug.Log("FIREED");
```
Fueredoriku: Remove pls

Fueredoriku (approved):
Very nice! 😁

PR #677, opened 2024-05-31
New shader for cratertown terrain

### LINKED ISSUE #518: Noise texture for terrain on crater town
Should only have 3 tones per color, and not have visible tiling

ChrisKrane:
Now with more sand and rocks!
Fix #518
image
image

### DIFF
--- a/Assets/Materials/CraterTownGround.mat
+++ b/Assets/Materials/CraterTownGround.mat
@@ -25,10 +25,11 @@ Material:
   m_Parent: {fileID: 0}
   m_ModifiedSerializedProperties: 0
   m_ValidKeywords:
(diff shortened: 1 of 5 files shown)


toberge (approved):
omg it looks so good! 🎉 

tried applying it to the menu terrain, but here the repetition becomes too obvious so I won't ask you to use this material there (yet):

image

ChrisKrane:
> omg it looks so good! 🎉
> 
> tried applying it to the menu terrain, but here the repetition becomes too obvious so I won't ask you to use this material there (yet):
> 
> image

I tried it too, and there are actually two materials for this (for some reason?), but yes the pattern is too obvious either way. Am considering making an adjusted version just for the menu terrain.

PR #678, opened 2024-05-31
Fix auction starting before all players spawned

toberge:
I may have adjusted the auction animation timing as well >:)


Fueredoriku (approved):
Looks good! 😎

PR #680, opened 2024-05-31
Add rail bridge and mcmarty sign, add tighter map borders

Fueredoriku:
More models will be added in a later PR (tm)


ChrisKrane:
The bridges look amazing!  However, it seems like the wrong terrain material has been applied, it should be the one called "Material/CraterTownTerrain".

toberge (approved):

PR #682, opened 2024-06-02
Make it possible to remove inputs

toberge:


toberge:
This will cause way too many merge conflicts right now, and I'm not sure we'll be able to do this properly

PR #685, opened 2024-06-02
Show loading screen when joining a lobby

toberge:
- [x] Show loading screen when joining a lobby until you receive info about other players in the lobby
- [x] Make loading screen cover all other UI (sort order 1000)
- [x] Disable sun in client lobby


Assets/Scenes/Menu.unity
Fueredoriku, line 1: Pain

Fueredoriku (approved):
Very nice 👍

PR #689, opened 2024-06-02
Perform a handshake before starting a round so that we know that all clients have initialized their players

toberge:
Has not seemed to cause that many issues yet, but there's bound to be some desynced mess here


toberge:
See #884

PR #690, opened 2024-06-02
Add orbit camera

### LINKED ISSUE #696: Option to spectate players in addition to map circling after death

toberge:
- [x] Orbit camera
- [x] ⚠️ Reset orientation of playerinput again smh
- [x] Adjust radius of the orbit by raycasting for obstacles (see that talk on how Journey did it maybe, lerping is your friend)
- [x] Make this thingy orbit other players after ragdoll shot (e.g. press space to spectate, enable/disable orbit cam and swap between living players)

Closes #696

### DIFF
--- a/Assets/Prefabs/Input/Player.prefab
+++ b/Assets/Prefabs/Input/Player.prefab
@@ -20,6 +20,7 @@ GameObject:
   - component: {fileID: 435425937908780799}
   - component: {fileID: 4693557610844660152}
   - component: {fileID: 8692423856261334150}
+  - component: {fileID: 4961168887133630009}
   m_Layer: 0
   m_Name: Player
   m_TagString: Untagged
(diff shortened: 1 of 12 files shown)


Fueredoriku (commented):
Some feedback:
- I think we still should keep the old "Killed by Player" while watching your own ragdoll, the text seems to have disappeared. (We could probably make the text a bit smaller and anchored at the top of the screen though while we're at it)
- We should also map "Fire" to change which player we are spectating
- I think it would be neat to display "Spectating playername" at the top of the screen when spectating a player

toberge:
Extra tidbits:
- perhaps a "press <key> to join" prompt should appear once we time out the death cam

Fueredoriku (approved):
Very nice!! Really makes the spectating so much better too 😁👍

PR #693, opened 2024-06-03
Avoid doubly loading menu

toberge:
Removed LoadSceneAsync calls that are performed *for us* by Mirror and thus resulted in some noticeably stutter when returning to the menu from
- the tutorial
- the end of a match
- otherwise disconnecting


Fueredoriku (approved):
Good find!

PR #699, opened 2024-06-04
Adjust ground threshold so certain ramps can be jumped off

toberge:


Fueredoriku (approved):
Feels so nice to not get stuck anymore!

PR #700, opened 2024-06-04
Restrict lobby joining by capacity ++

### LINKED ISSUE #659: Refactor lobby info in SteamManager
Regarding the `Lobbies` dict in SteamManager:

Ideally the string value should be an object with s'more interesting values (and the public access could be the .Values of the dict if the key was part of that dict, and the dict key could be the ulong value that is encapsulated in the CSteamID), but I'll let that slide for now as this menu needs a refactor for player count and so on later.

toberge:
- [x] Restrict lobby joining
- [x] Refactor lobby info, closes #659 
- [x] Fix steam name issue (now determined based on supplied steam ID)

### DIFF
--- a/Assets/Scripts/Utils/SteamManager.cs
+++ b/Assets/Scripts/Utils/SteamManager.cs
@@ -2,6 +2,8 @@ using System.Collections.Generic;
 using UnityEngine;
 using Mirror;
 using Steamworks;
+using System.Collections.ObjectModel;
+using UnityEngine.InputSystem.Controls;
 
 public enum AchievementType
 {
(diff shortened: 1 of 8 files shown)


Assets/Scripts/Utils/SteamManager.cs
```diff
@@ -164,16 +198,42 @@ private void OnlobbyCreated(LobbyCreated_t callback)
...
+    private void SetAsJoinable()
+    {
+        SteamMatchmaking.SetLobbyJoinable(lobbyID, true);
     }
 
     private void OnJoinRequest(GameLobbyJoinRequested_t callback)
     {
         // TODO verify that the lobby *should* be joined by more players!
         //      and verify that this is run on the server!
+        Debug.Log("LOBBY JOIN REQUEST");
```
Fueredoriku: I think we can remove this, also isn't the TODO now technically done? 🤔

Fueredoriku (approved):
Great stuff! 😁👍

Really like the refactor with the Lobby class!

Assets/Scripts/Utils/SteamManager.cs
toberge: Seems to work now, though I'm not sure if this method does *exactly* what I think it does.

PR #701, opened 2024-06-05
Big McMartys scrapyard auction and BBQ

Fueredoriku:
- Add tighter borders to cratertown
- Add Big McMartys scrapyard auction and BBQ along new cratertown borders
- Improved level design in cratertown
   - Made the one spawn with two chips less good by removing one of the chips
   - Opened up the same area a bit more, with added obstacles


toberge (approved):
looks very bigmcmarthy to me 👍

PR #703, opened 2024-06-05
Add private lobbies

toberge:
@Fueredoriku your move >:)


toberge:
Probably fixed in #747

PR #706, opened 2024-06-06
Fix crosshair not reseting HUD crosshair to default

### LINKED ISSUE #705: Rubber sniper sometimes locked in flaccid position
Screenshot_20240606_111236_Steam
Bug has been reproduced, seems to be to aimvector not always properly reseting after match end

Fueredoriku:
Resolves #705


toberge (approved):
Should probably happen in some reinitialization method (OnEnable/OnDisable) too, once we have implemented respawning, but this should fix the current issue 👍

PR #707, opened 2024-06-06
Fix #704

### LINKED ISSUE #704: Killing with shotguns grants 1 kill bonus per bullet
In other words, one can get 30 chips for 1 kill, yikes.

Fueredoriku:


Assets/Scripts/Gamestate/PlayerManager.cs
```diff
@@ -191,7 +191,8 @@ private void OnDeath(HealthController healthController, float damage, DamageInfo
         {
             killer = lastPlayerThatHitMe;
         }
-        onDeath?.Invoke(killer, this, info);
+        if (isAlive)
+            onDeath?.Invoke(killer, this, info);
```
toberge: Note that by guarding only this invocation, you're not preventing the ridiculous amount of knockback (and multiple invocations of the spectator camera code) that the shotgun can cause. I'd suggest turning this into an early return instead.

Fueredoriku: I agree that it should return, but the knockback issue stems from somewhere 
different (it also happens when you shoot an explosive barrel with a shotgun and get more knockback while still living)

Assets/Scripts/Augment/BulletModifiers/KnockbackOnShotModifier.cs
```diff
@@ -33,7 +36,10 @@ public void Detach(ProjectileController projectile)
     private void KnockAwayTargets(HitboxController controller, ref ProjectileState state)
     {
         Vector3 normal = gunController.transform.forward;
-
+        if (collidersHitWithShot.Item1 == state && collidersHitWithShot.Item2.Contains(controller))
```
toberge, line 39: note that you can actually name the fields in a tuple like `(Foo foo, Bar bar)`

Assets/Scripts/Augment/BulletModifiers/KnockbackOnShotModifier.cs
```diff
@@ -33,7 +36,10 @@ public void Detach(ProjectileController projectile)
...
+            return;
+        collidersHitWithShot.Item1 = state;
+        collidersHitWithShot.Item2.Add(controller);
```
toberge, line 48: When do we clear the list?

toberge (approved):
Seems good, though the list of hit colliders in knockbackmodifier seems to be able to grow indefinitely, perhaps it should be a set or something?

Assets/Scripts/Augment/BulletModifiers/KnockbackOnShotModifier.cs
Fueredoriku: Ye, slight oversight, will fix

Assets/Scripts/Gamestate/PlayerManager.cs
toberge: Oh dear.

Assets/Scripts/Augment/BulletModifiers/KnockbackOnShotModifier.cs
toberge: Also I'm not sure why this isn't just two variables?

Fueredoriku: Could have been, but I think they should be a tupple as they are only meant to be used together

PR #708, opened 2024-06-07
Reset match properly

toberge:
Should fix the wins-being-remembered issue if I've understood how to reproduce it correctly


Fueredoriku (approved):
There might be other fields we also have to reset in order to not see very wierd bugs, but the rounds field is probably the most important, so we can take the others later

PR #709, opened 2024-06-07
Reset camera rotation/position when spawning players

toberge:
This should fix the issue some players have where it does not reset after having been an orbiting spectator camera

(also reminds me that the spawn code needs refactoring, but that's not happening *right now*)


Fueredoriku (approved):
Have not tested, but the code seems like it should do the trick

toberge:
Actually doesn't properly fix the rotation 🌵

PR #710, opened 2024-06-07
Downscale HUD when only one local input

toberge:
image

to just a wee bit larger than for splitscreen


Fueredoriku (approved):
Looks better!

PR #711, opened 2024-06-08
Letterbox intro video

toberge:
Introvideo now keeps aspect ratio on all screen configurations, and has black bars instead of bleedthrough of the menu scene surrounding it. Had to add yet another camera for this smh

image


Fueredoriku (approved):
Nice change!

PR #712, opened 2024-06-08
Fix nullref exception in knockback on shot

toberge:
Can't clear an uninitialized list 💀
Also use named tuple members


Fueredoriku (approved):
Good fix

Issue #638, opened 2024-05-11
Victory screen (?) 


toberge:
Adding this idea as I wanna model instruments and pose auctioneer + bots with them for OST album cover, and thought this could be animated and used for a scene in the game.

toberge:
Solved in #805

PR #717, opened 2024-06-08
Add instruments >:)

### LINKED ISSUE #716: Instruments for a full Wild Metal Band >:)
# Models

- [x] Trumpet
- [x] Trumpetbot
- [x] Hornbot (@Fueredoriku)
- [x] Saxophone
- [x] Banjo
- [x] Pedal organ/piano
- [x] Drums
  - [x] Barrels (explosive and regular) for base/toms
  - [x] Cymbals
  - ~~Drumsticks~~ no need when hammer

# Posing

- [x] McMarty the drummer
- [x] Cowbot sax
- [x] Cowbot piano
- [x] Cowbot banjo
- [x] Curiosity trumpet (no animation, it's dead)
- [x] Trumpetbot trumpet
- [x] Hornbot horn

# Stage

- [x] Extract the bit of crater town that has McMarty's auction house
- [x] Plonk down the band
- [x] Set the scene

toberge:
# wth does this do

- Closes #716 and thus adds a bunch of new props and animations 🎉 
- Adds instrumentbots to crater town and the pit, shooting them distorts the music
- Adds instrumentbots and a piano player to the main menu
- Adds a piano to the auction stage
- Adds a scene with the full band and a render camera for the album cover
- Adds a script for writing render textures to file

# Future plans/needs

- Fix color encoding in render texture writing script
- Adjust animation speed to follow the beat of the currently playing song (for horn- and trumpetbot only)
- Add victory screen with band members and some pedestals and stuff, see #638


Assets/Scripts/Utils/RenderTextureActions.cs
```diff
@@ -0,0 +1,32 @@
...
+        AssetDatabase.ImportAsset(path);
+        Debug.Log("Saved to " + path);
+    }
+
+    [MenuItem("Assets/Save RenderTexture to file", true)]
+    public static bool ValidateAsRenderTexture()
+    {
+        return Selection.activeObject is RenderTexture;
+    }
+}
```
Fueredoriku: Newline pls

Fueredoriku (approved):
The band is very cute, and it's so funny that we finally added this 😁

I have some general feedback, but there is nothing major, so I'll leave an approve for you to use at your discretion 👍

General feedback:
- I think the band could use a name! Maybe add a loadingscreen tip where you say something like "did you know the music is played by *insert band name here* "
- The hornbot animation no longer produces particles. It seems that maybe the animation may have lost the "set active"-toggles of the gameobjects that contain the exhaust particles? 🤔
- For the main menu, while the scales technically are correct, I think adjusting the size or placement of the piano+cowbot so the size looks more comparatively correct to the barrels (which look much larger than they actually are in game).
image

PR #718, opened 2024-06-08
Prevent light from leaking through menu terrain

toberge:
Jesper's suggestion seems to work:
Place a plane under the terrain that blocks the light from the sun and prevents the odd lighting that would appear at night.


Fueredoriku (approved):
Nice clean fix!

PR #720, opened 2024-06-09
Move the ammobox that was too high down

toberge:
The child transform was moved, making it appear higher up than it should be

it's this one: (corrected version)
image

~~now that I think about it, didn't I actually see this w/o the ammo box body? I thought I had the dancer~~ 🤔 
nvm, forgot that I was spectating the bots


Fueredoriku (approved):
Nice!

PR #722, opened 2024-06-09
Provide correct aim orientation when shooting

### LINKED ISSUE #639: Synchronize where the gun fires from (and in which direction it does)

toberge:
Should fix *some* of the issues we experience with the current netcode, but we still need to implement rollback of some kind.

Closes #639 

should test like all the gunparts, but only trash uplink seemed to have real issues with this


Fueredoriku (approved):
Great stuff, really makes a lot of difference for non-host players 👍

PR #725, opened 2024-06-10
Fix #721, add new coordinates in scene for better floor shadows

### LINKED ISSUE #721: Shake-on-overbid animation cancels intro animation and leaves the camera in a bad state
image

Fueredoriku:
Closes #721, also changes floor hatched shadows to single lines instead of cross hatching


toberge (approved):
Code looks alright, though float comparisons are always a little scary. Haven't tested as I'm quite busy >:)

PR #726, opened 2024-06-15
Increase coil barrel accuracy slightly

toberge:


Fueredoriku (approved):
Might be a bit strong now, but it's always nice to tone down attributes that can be experienced as a bit frustrating 😅

We'll hav to test it a bit, but I think this is fine 👍

PR #736, opened 2024-06-25
Give lazur shots larger radius

toberge:
Previous radius: 0

This should make the lazur barrel much more viable

The way I've upped the radius might lead to some odd edgecases, and this also needs testing


Fueredoriku (approved):
A subtle, but much needed buff! Nice that multi-bounce actually works now (even though it still is a bit hard to perform) 👍

PR #737, opened 2024-06-25
wip fix reload buggerino

toberge:


toberge:
Current theory: this is ~~all~~ partly due to the netcode making firing the gun cause a round trip to the server. should be instant when running locally, but seems to not be so.

toberge:
Resolved in another pull request a while ago.

PR #738, opened 2024-06-25
Add straight bullet paths with telescope, buff telescope stats

### LINKED ISSUE #446: Zooming extension rework
Giant crosshair should be removed based on feedback

Could us a charge-up head shot mechanic as well to be more useful

Fueredoriku:
Should make telescope more interesting and unique as an augment
- Telescope now makes all bullets travel straight with no spread or gravity.
- Projectile speed is also slightly increased


toberge (approved):
This makes the telescope much more impactful, very nice 👍 

Do note that the projectile speed increase affectst the cannon heavily and makes e.g. sodie popper + telescope perhaps _overly_ good with the cannon. We may need to adjust how impactful the speed factor is on the cannon -- but perhaps later, as it is already severely powerful with the portal extension.

PR #740, opened 2024-06-26
Fix navmesh in "The Pit", fix AIs getting stuck

Fueredoriku:
AIs get stuck way less frequently now 🤠
Fixed:
- AIs getting stuck in fans
- AIs getting stuck in the air due to knockback
- AIs getting softlocked in a close up stand off where they can't hit eachother


toberge (approved):
Seems to work well (now that I fixed it)

PR #741, opened 2024-06-26
Cursor support and sensitivity setting!

CraZyB1336:
Cursor now has the ability to select buttons and other UI items in the main menu.
Sensitivity setting is now adjustable with an input field and a slider.
Same with FOV and Zoom FOV


Assets/Scripts/Control&Input/PlayerInputManagerController.cs
```diff
@@ -107,6 +109,9 @@ public void ChangeInputMaps(string mapNameOrId)
     {
         LocalPlayerInputs.ForEach(playerInput =>
         {
+            Cursor.visible = playerInput.IsMouseAndKeyboard && mapNameOrId.Equals("Menu");
```
toberge: I'm not sure why this works? Shouldn't this cause the cursor to be invisible if the last input is a gamepad?

I'd suggest writing it as
```csharp
Cursor.visible = LocalPlayerInputs.Any(i => i.IsMouseAndKeyboard) && mapNameOrId.Equals("Menu");
```
before the ForEach.

Assets/Scripts/Control&Input/PlayerMovement.cs
```diff
@@ -37,6 +37,8 @@ public class PlayerMovement : MonoBehaviour
     [SerializeField]
     public float LookSpeedZoom = 0.75f;
 
+    private float sensScale;
```
toberge, line 40: Consider writing unabbreviated here, as `sens` may not be entirely intuitive.

Assets/Scripts/Control&Input/PlayerMovement.cs
```diff
@@ -181,12 +183,23 @@ public void SetPlayerInput(InputManager player)
...
             MatchController.Singleton.onRoundEnd += ResetZoom;
     }
 
+    public void SetFOVFromSettings()
+    {
+        if (playerCamera != null)
+        {
+            startingFov = SettingsInfo.Singleton.settings.playerFOV;
+            playerCamera.fieldOfView = SettingsInfo.Singleton.settings.playerFOV;
+            ZoomFov = SettingsInfo.Singleton.settings.zoomFOV;
```
toberge: Note that this is overridden by the telescope extension. Should be refactored to be a factor, probably, so that you can do `ZoomFovFactor * theActualFovFromSettings`

Assets/Scripts/SettingsInfo.cs
```diff
@@ -0,0 +1,253 @@
...
+    public int resolutionPresetIndex;
+    public int displayMode;
+
+    // Gameplay
+    public float sensScale;
+    public float playerFOV;
+    public float zoomFOV;
+}
+
+public class SettingsInfo : MonoBehaviour
```
toberge: Not really a fan of the `Info` suffix, but ignore this :))))))

Assets/Scripts/SettingsInfo.cs
```diff
@@ -0,0 +1,253 @@
...
+    [Header("FOV Limits")]
+    public float lowerFOVLimit = 60f;
+    public float upperFOVLimit = 120f;
+
+    [Header("Zoom FOV Limits")]
+    public float lowerZoomFOVLimit = 20f;
+    public float upperZoomFOVLimit = 50f;
+
+    #endregion
+    public Settings settings = new()
```
toberge: Please capitalize public variables :pray:

Assets/Scripts/SettingsInfo.cs
```diff
@@ -0,0 +1,253 @@
...
+    public float musicVolume;
+    public float sfxVolume;
+
+    // Graphics
+    public int qualityPresetIndex;
+    public int resolutionPresetIndex;
+    public int displayMode;
+
+    // Gameplay
+    public float sensScale;
```
toberge: sensitivity, please :upside_down_face:

Assets/Scripts/SettingsInfo.cs
```diff
@@ -0,0 +1,253 @@
...
+public struct Settings
+{
+    // Audio
+    public float masterVolume;
+    public float musicVolume;
+    public float sfxVolume;
+
+    // Graphics
+    public int qualityPresetIndex;
+    public int resolutionPresetIndex;
```
toberge: This *may* break if the user changes monitor

Assets/Scripts/SettingsInfo.cs
```diff
@@ -0,0 +1,253 @@
...
+
+    private const string FileName = "/Settings.json";
+
+    private static string FilePath;
+    #region Graphics variables
+    public Resolution[] resolutions { get; private set; }
+
+    public string[] qualityNames { get; private set; }
+    #endregion
+    #region Audio variables
```
toberge: Please put a line or two of whitespace between regions

Assets/Scripts/SettingsInfo.cs
```diff
@@ -0,0 +1,253 @@
...
+    public float upperZoomFOVLimit = 50f;
+
+    #endregion
+    public Settings settings = new()
+    {
+        sensScale = 1.0f,
+        masterVolume = 1.0f,
+        musicVolume = 1.0f,
+        sfxVolume = 1.0f,
+        qualityPresetIndex = 0,
```
toberge: Shouldn't this be Ultra by default? :thinking:

Assets/Scripts/SettingsInfo.cs
```diff
@@ -0,0 +1,253 @@
...
+
+    private void Awake()
+    {
+        resolutions = Screen.resolutions.Reverse().ToArray();
+        qualityNames = QualitySettings.names;
+
+        mainAudioMixer.GetFloat(audioGroupMusic, out float musicVolume);
+        maxVolumeMusic = musicVolume;
+        mainAudioMixer.GetFloat(audioGroupSFX, out float sfxVolume);
+        maxVolumeSFX = sfxVolume;
```
toberge: Hm, I wonder why the master volume wasn't treated the same way?

And for that matter, I'm not sure why this was necessary in the first place?

Assets/Scripts/SettingsInfo.cs
```diff
@@ -0,0 +1,253 @@
...
+
+        // Check lower limits for Gameplay
+        if (settings.sensScale <= 0f)
+            settings.sensScale = 0.01f;
+
+        if (settings.playerFOV < 20f)
+            settings.playerFOV = 20f;
+
+        if (settings.zoomFOV < 5f)
+            settings.zoomFOV = 5f;
```
toberge: These could be replaced with `settings.zoomFOV = Mathf.Max(5f, settings.zoomFOV)` if you want :))))

Assets/Scripts/SettingsInfo.cs
```diff
@@ -0,0 +1,253 @@
...
+    }
+    #endregion
+    #region Gameplay methods
+    public float ClampSensValue(float value)
+    {
+        return Mathf.Clamp(value, lowerSensLimit, upperSensLimit);
+    }
+    public void SetSensMultiplier(float scale)
+    {
+        settings.sensScale = scale;
```
toberge: Huh. I'm not sure why you're somehow able to mutate a struct here. This *works*, but the fact that something like this actually works seems rather fishy – structs are supposed to be immutable :thinking:

Assets/Scripts/UI/MainMenu/MainMenuController.cs
```diff
@@ -111,10 +119,20 @@ private void Start()
             mainMenuCamera.SetActive(true);
             // Reset loading screen
             LoadingScreen.ResetCounter();
+
+
+            if (firstInputJoined.IsMouseAndKeyboard) EnableVisibleMouse();
```
toberge: Basically the same code in EnableVisibleMouse as in the `else` branch, consider making the `else` branch call its own method

Assets/Scripts/UI/MainMenu/MainMenuController.cs
```diff
@@ -111,10 +119,20 @@ private void Start()
...
+            {
+                Cursor.lockState = CursorLockMode.Locked;
+                Cursor.visible = false;
+            }
         }
         else
         {
             // First time in menu, play intro video.
+            Cursor.lockState = CursorLockMode.Locked;
+            Cursor.visible = false;
```
toberge: and then call that method here as well

Assets/Scripts/UI/MainMenu/MainMenuController.cs
```diff
@@ -281,6 +318,12 @@ private void AddPlayer(InputManager inputManager)
         levelSelectManager.SetPlayerInput(inputManager);
     }
 
+    private void EnableVisibleMouse()
```
toberge: ShowMouse()?

Assets/Scripts/UI/MainMenu/MainMenuController.cs
```diff
@@ -205,6 +224,21 @@ private IEnumerator WaitSelect(Selectable target)
         target?.Select();
     }
 
+    public void DeselectControl()
+    {
+        if (EventSystem.current != null)
+        {
+            // I find no better way to deselect game objects. This is mainly used so you don't have to click twice in tabs.
+            StartCoroutine(WaitUnselect());
```
toberge, line 227: Seems fair considering the usual Unity problems :pensive:

Assets/Scripts/UI/MainMenu/OptionsMenu.cs
```diff
@@ -25,36 +26,87 @@ public class OptionsMenu : MonoBehaviour
...
+        sfxVolumeSlider.value = settingsInfo.settings.sfxVolume;
+
+        sensitivitySlider.value = settingsInfo.settings.sensScale;
+        sensitivityInputField.text = settingsInfo.settings.sensScale.ToString("0.00");
+
+        FOVSlider.value = settingsInfo.settings.playerFOV;
+        FOVInputField.text = settingsInfo.settings.playerFOV.ToString("0.00");
+
+        ZoomFOVSlider.value = settingsInfo.settings.zoomFOV;
+        ZoomFOVInputField.text = settingsInfo.settings.zoomFOV.ToString("0.00");
```
toberge: Could we format the FOV values as whole numbers, please? The .xx is a little distracting when it likely doesn't matter much.

Assets/Scripts/UI/MainMenu/OptionsMenu.cs
```diff
@@ -25,36 +26,87 @@ public class OptionsMenu : MonoBehaviour
...
+        qualityDropdown.value = settingsInfo.settings.qualityPresetIndex;
+
+        masterVolumeSlider.value = settingsInfo.settings.masterVolume;
+
+        musicVolumeSlider.value = settingsInfo.settings.musicVolume;
+
+        sfxVolumeSlider.value = settingsInfo.settings.sfxVolume;
+
+        sensitivitySlider.value = settingsInfo.settings.sensScale;
+        sensitivityInputField.text = settingsInfo.settings.sensScale.ToString("0.00");
```
toberge: Perhaps only one digit of precision?

Assets/Scripts/UI/MainMenu/OptionsMenu.cs
```diff
@@ -63,45 +115,94 @@ private void CheckResolutions()
...
 
     public void SetDisplayMode(int displayMode)
     {
-        Screen.fullScreenMode = (FullScreenMode)displayMode;
+        settingsInfo.SetDisplayMode(displayMode);
+    }
+
+    public void CheckSensInputValue(string inputValue)
+    {
+        sensitivityInputField.text = settingsInfo.ClampSensValue(float.Parse(inputValue)).ToString("0.00");
```
toberge: Why not do this as part of the `Change` methods?

Assets/Scripts/SettingsInfo.cs
```diff
@@ -0,0 +1,253 @@
...
+        CreateDefaultFile();
+    }
+
+    public void ApplyAllSettings()
+    {
+        SetResolutionLevel(settings.resolutionPresetIndex);
+        SetDisplayMode(settings.resolutionPresetIndex);
+        SetQualityLevel(settings.qualityPresetIndex);
+        SetMasterVolume(settings.masterVolume);
+        SetMusicVolume(settings.musicVolume);
```
toberge: What about SFX volume? :)))))

Assets/Scripts/UI/MainMenu/OptionsMenu.cs
```diff
@@ -63,45 +115,94 @@ private void CheckResolutions()
...
-
     public void SetMasterVolume(float volume)
     {
-        mainAudioMixer.SetFloat(audioGroupMaster, LinearToLogarithmicVolume(volume));
+        settingsInfo.SetMasterVolume(volume);
     }
 
     public void SetMusicVolume(float volume)
     {
-        mainAudioMixer.SetFloat(audioGroupMusic, LinearToLogarithmicVolume(volume) + maxVolumeMusic);
```
toberge, line 84: I'm not sure what the `+ maxVolumeMusic` thing is all about, seems suspicious

toberge (changes requested):
Questionable quirks:

- Volume seems to be multiplied by the volume setting each time you enter the main menu, resulting in gradually lower volume if you've set it < 1
- The map cards are oddly hard to select with the mouse, it seems like the selection is stuck on the first item until you hover over the one that is selected and then try again?
- Focusing on a tab causes focus to effectively be lost if you try to navigate using keyboard from there

Other than that, it's real nice to have some proper options and be able to use the mouse in the menu.

Assets/Scripts/SettingsInfo.cs
Fueredoriku: FYI, this was added to prevent fading between music tracks to fade from max volume instead of the volume level the settings use, but it might be possible to just do this on master volume only

Assets/Scripts/UI/MainMenu/OptionsMenu.cs
toberge: After discussing with @Fueredoriku, this should be replaced with a constant based on the current max volume (in the log translation function itself)

CraZyB1336:
> Questionable quirks:

>     * The map cards are oddly hard to select with the mouse, it seems like the selection is stuck on the first item until you hover over the one that is selected and then try again?

There seems to be some strange hitbox behavior with the cards and the EvenSystem's raycaster. I'll experiment a little further on this issue

Assets/Scripts/Control&Input/InputManager.cs
```diff
@@ -100,11 +103,9 @@ public void AddListeners()
         playerInput.actions["Look"].performed += Look;
         playerInput.actions["Look"].canceled += Look;
 
-        // Imprison mouse
```
Fueredoriku, line 103: This is so sad 😢

toberge (approved):
LGTM :+1:

PR #744, opened 2024-06-27
Add basic LOD

### LINKED ISSUE #735: Set up LOD
Adding LOD for our models would not only give us better performance, but would also improve the outline graphics, as we would have less hard edges from a distance. This way objects don't become jumbled black messes at medium distances.
2 for 1!

Objects that should be LOD-ed:
- [x] Players
- [x] ExplodingBarrels

Fueredoriku:
Resolves #735

Have done:
- Explosive barrels
- Players


toberge (approved):
LGTM, thanks for going through the pain soldier 🫡

PR #746, opened 2024-06-28
Fix item select timeout

toberge:
Also adjust error margin for input


Fueredoriku (approved):
Nice to get this fixed 👍

PR #747, opened 2024-06-30
Add simple lobby queue

Fueredoriku:
Could be improved with its own loading screen and several fetching attempts instead of just one before creating a new lobby, but I think this will work better than the current lobby system for now as is.

### DIFF
--- a/Assets/Prefabs/UI/Menus/LobbySelect.prefab
+++ b/Assets/Prefabs/UI/Menus/LobbySelect.prefab
@@ -104,9 +104,9 @@ RectTransform:
   m_LocalScale: {x: 1, y: 1, z: 1}
   m_ConstrainProportionsScale: 0
   m_Children:
-  - {fileID: 1656343962051143921}
-  - {fileID: 8029557614816138824}
   - {fileID: 4792437529069775652}
+  - {fileID: 6291501560300597862}
+  - {fileID: 1656343962051143921}
   - {fileID: 498079454081798691}
   m_Father: {fileID: 7055458770637592391}
   m_RootOrder: -1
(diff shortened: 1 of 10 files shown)


Assets/Scripts/Gamestate/Ruleset.cs
```diff
@@ -163,6 +163,14 @@ public struct NetworkRuleset
         };
 }
 
+public enum LobbyGameMode
```
toberge: I'd rather call it `GameModeType` or something, so that it isn't *only* tied to lobbies

Assets/Scripts/UI/MainMenu/QueueMenu.cs
```diff
@@ -0,0 +1,34 @@
...
+    }
+    private void SetUpQueue()
+    {
+        var viableLobbies = SteamManager.Singleton.Lobbies
+            .Where(lobby => lobby.gameMode == MatchRules.Singleton.Rules.GameMode)
+            .OrderBy(lobby => lobby.availableSlots);
+
+        if (viableLobbies.ToArray().Length > 1)
+        {
+            var enumerator = viableLobbies.GetEnumerator();
```
toberge, line 25: double enumeration, consider ToList()ing it.

Assets/Scripts/UI/MainMenu/QueueMenu.cs
```diff
@@ -0,0 +1,34 @@
...
+            do
+                if (SteamManager.Singleton.RequestLobbyJoin(viableLobbies.First().id))
+                    return;
+            while (enumerator.MoveNext());
+        }
+
+        // No matching lobbies found, creating new
```
toberge: We should ideally show this as a message to the user

Assets/Scripts/Utils/SteamManager.cs
```diff
@@ -42,16 +42,20 @@ public void UpdateMetadata()
         availableSlots = 0;
         if (int.TryParse(SteamMatchmaking.GetLobbyData(id, AvailableSlotsProperty), out var slots))
             availableSlots = slots;
+        if (int.TryParse(SteamMatchmaking.GetLobbyData(id, GameModeProperty), out var type))
+            gameMode = (LobbyGameMode)type;
     }
 
     public static string NameProperty = "name";
     public static string AvailableSlotsProperty = "availableSlots";
+    public static string GameModeProperty = "gameMode";
```
toberge, line 51: We should also have a property for the version of the game on the host's side. Could simply use `Application.version`, as long as we make sure to update that when we push new stuff to the beta branch.

Assets/Scripts/Utils/SteamManager.cs
```diff
@@ -314,6 +319,7 @@ public void LeaveLobby()
 
     public void FetchLobbyInfo()
     {
+        isQueueing = false;
         FetchFriendLobbyInfo();
         // Ensure we have enough available slots
         SteamMatchmaking.AddRequestLobbyListNumericalFilter(Lobby.AvailableSlotsProperty,
```
toberge, line 329: There's also filtering in the `FetchFIlteredLobbyInfo` method, what's the distinction here?

Assets/Prefabs/UI/Menus/LobbySelect.prefab
```diff
@@ -970,7 +731,7 @@ PrefabInstance:
       objectReference: {fileID: 0}
     - target: {fileID: 7072107538278606930, guid: e472bbf090cca566cae93eec176f8dff, type: 3}
       propertyPath: m_Name
-      value: Local
+      value: Private
```
toberge: This isn't a private lobby, it's a local lobby currently. Could we leave in both private and local lobbies, please? Cuz if private means friends only then they'll be able to hop in.

toberge (changes requested):
Doesn't work at the moment. Just doesn't connect to steam somehow 🤔 
image

Current bugs:
- needs to set lobby mode in MainMenuController in order for StartHost to work
- shouldn't even call StartHost though
- actually, the filtering method in SteamManager isn't called from anywhere?
- even adding that didn't resolve issues

Concerns:
- I'm certain that calling local-only lobbies _private_ will confuse people, as they can't invite friends to them
- Calling friend-only lobbies private also seems a lil weird
- The choose-a-lobby UI could see some extended life for joining friend lobbies, as in that case players may want to select a specific friend's lobby

Assets/Scripts/UI/MainMenu/QueueMenu.cs
Fueredoriku: I can debug it out, but creating an in game log with feedback for users goes into the issue-pile

Assets/Scripts/Gamestate/Ruleset.cs
Fueredoriku: Fixed

Assets/Scripts/UI/MainMenu/QueueMenu.cs
Fueredoriku: Changed it up the way I originally intended

Assets/Scripts/Utils/SteamManager.cs
Fueredoriku: Fixed, also incremented version number 👍

Assets/Scripts/Utils/SteamManager.cs
Fueredoriku: Renamed

Assets/Prefabs/UI/Menus/LobbySelect.prefab
Fueredoriku: Now we have queue, friends only and local.

Assets/Scripts/Utils/SteamManager.cs
toberge: Um... I'm sorry, but I upped the version number for this release already (last release was 0.3.something)

Assets/Scripts/UI/MainMenu/QueueMenu.cs
toberge: Okay we'll do it later ™️ 
I just thought it could be like "searching for open lobbies..." and then "no lobbies found, hosting new lobby" in a loading screen or smth

Assets/Scripts/UI/MainMenu/QueueMenu.cs
toberge: Don't see why you did it as a `while` with GetEnumerator and not as a `foreach`, but this looks fine

Assets/Scripts/Utils/SteamManager.cs
toberge: Ah, so this is going to be for cases where we don't filter by gamemode. Could perhaps make the name of _this_ method a little clearer as well, but this seems fine enough.

toberge (approved):
Works well now, you just gotta fix the version number >_>

Assets/Scripts/Utils/SteamManager.cs
Fueredoriku: Fixed 👍

PR #749, opened 2024-07-01
Fix CAN WIN appearing too early on any scoreboard, and coil+scope combo

### LINKED ISSUE #750: Coil + telescope does not shoot straight across network
Seems like the zigzag modifier isn't actually disabled unless it's the local player or fps gun?

edit: yep, the AIs don't even trigger it

toberge:
whoops

Closes #750


Fueredoriku (approved):
Looks good 👍

PR #754, opened 2024-07-04
Fix outline shader for non-standard materials, Add better explosion feedback

### LINKED ISSUE #739: More impactful explosions
Should at least produce:
- [x] Shockwave
- [x] Screen shake (if hit)

Fueredoriku:
- Fixed sobel detection of rewritten z buffer (allows outlines on jigglematerials and even mesh particles!!)
- Add pressure wave effect to explosive barrels for better feedback
- Fix issue with rope that could crash games in some weird edge cases


toberge (approved):
LGTM 👍

PR #755, opened 2024-07-05
Add subtle jump and a little less (but still) subtle leap audio

Fueredoriku:


toberge (approved):
Gonna leave an approve here. The sound triggers at the correct points, but it is almost inaudible for regular jumps, and still barely there for leaps.

Assets/Scripts/Gamestate/PlayerManager.cs
```diff
@@ -499,7 +519,7 @@ private void SetLayerOnSubtree(GameObject node, int layer)
 
     protected void PlayOnHit()
     {
-        if (Random.Range(0, 10000) > 5)
+        if (Random.Range(0, 100000) > 5)
```
toberge, line 522: 😔

PR #760, opened 2024-07-08
Fix Collider::ClosestPoint() warnings

toberge:
ClosestPoint() can only be used on some colliders, this should avoid most cases of that warning


Fueredoriku (approved):
Nice, less clutter in the log is always a plus 👍

PR #761, opened 2024-07-08
Add killfeed

### LINKED ISSUE #439: Killfeed
Feedback for player kills

toberge:
Closes #439 

max capacity 6

fades

has player colors

I ded

### DIFF
--- a/Assets/Scripts/Augment/ProjectileDamageController.cs
+++ b/Assets/Scripts/Augment/ProjectileDamageController.cs
@@ -6,9 +6,10 @@ public enum DamageType
     Explosion,
     Falling,
     Continuous,
+    // TODO emit these two damage types (or do a separate effect type?)
     Fire,
     Hacking,
-    // TODO determine if "knockback" should be its own thing
+    // TODO determine if "knockback" should be its own type (effect type?)
 }
 
 public struct DamageInfo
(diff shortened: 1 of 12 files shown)


Assets/Scripts/UI/EventLog.cs
```diff
@@ -0,0 +1,96 @@
...
+        while (itemsInUse.Count > 0 && Time.time - itemsInUse.Peek().creationTime > expiryTime)
+        {
+            var item = itemsInUse.Dequeue();
+            // Fade and put back in queue after tween is done.
+            LeanTween.value(item.text.gameObject, SetAlpha(item.text), 1, 0, .2f).setOnComplete(() =>
+            {
+                item.text.enabled = false;
+                item.text.transform.SetAsLastSibling();
+                availableItems.Enqueue(item.text);
+            });
```
Fueredoriku: There seems to be a bug where if enough items has been used, typically in the third shooting round, messages will no longer appear visually. I think it might be due to some of these shenaningans, specifically I think you need tween cancel guards that always properly resets the alpha of the canceled item so it's visible again the next time it's in use.

Fueredoriku (changes requested):
Very nice addition, and the differentiation between damage types is GREAT 😁👍

There seems to be a bug where if enough of the items in the eventlog has been used, their alpha is always 0 when reused, if you fix that I think this should be good to go

Assets/Scripts/UI/EventLog.cs
toberge: Uh oh, so that did turn out to be necessary -_-

toberge: The only required part of the fix was the `alpha = 1` reset when a new item is logged, but I went ahead and fixed the issue that *could* appear if we were to exceed the capacity of the log (I originally wrote it so that items were temporarily outside both queues while they faded out)

Assets/Scripts/Augment/ProjectileDamageController.cs
```diff
@@ -42,6 +43,32 @@ public DamageInfo(PlayerManager source, NetworkDamageInfo info)
...
+        : $"{sourcePlayer.identity.ToColorString()} {KillTypeToString()} {victim.identity.ToColorString()}";
+
+    private readonly string SuicideTypeToString() =>
+        damageType switch
+        {
+            DamageType.Explosion => "exploded themselves",
+            DamageType.Falling => "fell to their death",
+            // TODO fix the emitted damage type from fire extension
+            DamageType.Continuous => "burned in their own trail",
+            _ => "committed sudoku"
```
Fueredoriku, line 59: lol

Fueredoriku (approved):
Works as expected now 👍

PR #762, opened 2024-07-08
Fix bidding music not always starting on clients

toberge:
The method in MatchController wouldn't always be called smh


Fueredoriku (approved):
Nice to get this fixed 👍

PR #764, opened 2024-07-08
Fix solar body not using toon shader

Fueredoriku:


toberge (changes requested):
image

image

The solar panel doesn't have specular highlight anymore, is this intentional?

Fueredoriku:
Yes, it is intentional, It looks bad.

Might look better with a noisy normal map, but fixing the solar UV stuff is something I would rather do in a seperate PR

toberge (approved):
Aha. Then I approve :+1:
(and hopefully we figs later)

PR #765, opened 2024-07-09
Fix shadow texture hatching being dependent on world space

Fueredoriku:
Now works the same way at any world space coordinate


toberge (approved):
LGTM

Issue #731, opened 2024-06-21
Sticky bombs barrel (?)


toberge:
can attach to pan

PR #766, opened 2024-07-09
Nerf damage and projectile speed of several parts

toberge:
Controversy expected.

- Cannon barrel: Lower projectile and explosion damage
- Shotgun barrel: Lower damage per projectile
- DDR body: Lower projectile speed
- Telescope extension: Lower projectile speed

Couldn't nerf Seeker damage as we don't have stat modifiers for bodies yet (╯°□°）╯︵ ┻━┻


Fueredoriku:
> Cannon barrel: Lower projectile and explosion damage

Probably a good change, especially if we intend to make #731. (We open up the design space for a projectile with a harder to hit trajectory, but more damage)

> Shotgun barrel: Lower damage per projectile

I think lowering the amount of bullets could also be an alternative here, but only if it doesn't change the nice visuals and feedback the shotgun currently has. Which is better, nerfing damage or lowering bullets, is hard to say. Demo sampling needed I think 😅

> DDR body: Lower projectile speed

After testing I agree, the previous speed was a bit much and it still exhibits the "fastness" I intended originally, so this is fine imo

> Telescope extension: Lower projectile speed

This one I'm less sure about. The intention with the part as of current is to make your weapon like a sniper rifle, lowering the projectile speed makes this much less so. Alternatives if you want to nerf this part and keep the right feel for a sniper could be:
- Lowering the firerate (makes it feel like you have put a lot of force into the projectile and it needs to "recharge") 
- Only have the fast projectile speed while scoped (to keep the feel of fast precision aiming)

Fueredoriku (approved):
Nice changes 👍

PR #768, opened 2024-07-10
Fix overlay stencil for ONLY damagepopup instead of every stencil font

Fueredoriku:
Fixed issue where every instance of the stencil font would be overlayed over geometry and not just the damage popup

Now we have a seperate font for overlay, which we can possible use later for stopping transparent edge shader bleedthrough on UI


toberge (approved):
Seems to work, nice :+1:

PR #769, opened 2024-07-10
Add toonshader VFX material for meshparticles

Fueredoriku:
Also updated hat projectiles to look just like the real deal 🤠


toberge (approved):
it finally wrok

PR #777, opened 2024-07-11
Add yielding to auction, fix specular sign, fix LOD player color in auction

### LINKED ISSUE #229: Yieldable auction
When all players are done bidding, some wanted to yield in order to speed up the process. One could also autimaticaly speed up the timer when noone has more money to bid with

Fueredoriku:
image


Assets/Scripts/Auction/AuctionDriver.cs
```diff
@@ -115,6 +125,70 @@ private IEnumerator WaitAndStartCameraAnimation()
...
+    {
+        if (yieldingPlayers.Contains(player))
+            return;
+
+        yieldingPlayers.Add(player);
+        OnYieldChange?.Invoke();
+
+
+        if (isAuctionStart && IsAuctionYielded())
+            StartCoroutine(nameof(WaitAndTryAuctionEnd));
```
toberge, line 147: else stop the coroutine! 

currently it will
- wait for 3 seconds before checking if it should stop :skull: 
- show SKPPING IN X after going back to YIELD X/Y when someone leaves, until those 3 seconds have passed

toberge (approved):
Looks good, other than the coroutine that should be ended!

Assets/Scripts/Auction/AuctionDriver.cs
Fueredoriku: Fixed 👍

PR #778, opened 2024-07-11
Add shockwave to cannon explosions

toberge:
I felt this to be necessary, especially for showing the huge radius of the enlargened explosion more clearly.

Also made explosions deal at least 1 damage.


Fueredoriku (approved):
Good changes 👍

PR #781, opened 2024-07-12
Refactor player initialization

toberge:
Refactored to

1. Happen in PlayerFactory rather than the network manager
2. Avoid duplicate code
3. Perform identity update within the identity itself


Fueredoriku (approved):
Nice refactor 👍

PR #783, opened 2024-07-12
Fix annoying navmesh agent exceptions

toberge:


Assets/Scripts/Gamestate/AIManager.cs
```diff
@@ -249,7 +249,8 @@ private void FindPlayers()
         agent.stoppingDistance = ShootingTarget ? shootingStoppingDistance : itemStoppingDistance;
         try
         {
-            agent.SetDestination(DestinationTarget.position);
+            if (agent.enabled)
+                agent.SetDestination(DestinationTarget.position);
```
Fueredoriku, line 253: I think you need to add an 
`
else
{
   aimovement.enabled = true;
}
` to preserve the behavior here

Fueredoriku (approved):
Should be good if you resolve the comment 😁👍

PR #795, opened 2024-07-13
Prevent ragdoll cam from starting during scoreboard anim

### LINKED ISSUE #793: Camera can be locked on ragdoll instead of showing scoreboards at times

toberge:
Closes #793 

Also refactors so that inputmanager and camera can only be set by passing them to the Activate method. The previous iteration could result in invalid state.


Fueredoriku (approved):
Nice fix 👍

PR #797, opened 2024-07-13
Fix missing size parameter for gasoline explosion

Fueredoriku:
Removed the annoying size parameter missing excepetions


toberge (approved):
LGTM but so sad no implementation yet smh

PR #801, opened 2024-07-13
Fix persistent yield count, remove shadow texture from enlarger

Fueredoriku:


toberge (approved):
Yeah this seems to work :+1:

PR #803, opened 2024-07-14
Give AI players a budget to spend in auctions

### LINKED ISSUE #863: Limit AI player spending in auctions
They need a budget 😤

toberge:
- [x] define simple budget
- [x] make it depend on gamemode (only in first to 30 it takes effect)
- [x] make bots yield when they've reached budget
- [x] and prevent them from bidding on things that would take them beyond budget (unless especially good I suppose?)

Closes #863


Fueredoriku (approved):
Nice improvement of the first to 30 AIs!

It would be nice if they additionaly didn't bid on augments they already own too, but that could be done in a later PR, so I'll leave an approve for you to use at your discretion 👍

PR #805, opened 2024-08-04
Add victory screen

toberge:
- [x] Animate victory sweep
- [x] Replace win screen overlay with just text @ bottom that appears when we reach winner
- [x] Place player model at victory position
- [x] Set winner's color
- ~~Make winner strike a pose (or dance)~~
- [x] Add victory scenery to maps (just a bunch of band members and the victor)
- [x] Use animation events to trigger
  - win text
  - victory pose
  - back to main menu (implicitly, after victory pose)
- [x] Compose victory fanfare (start of wild metal band)
- [x] DRUM ROLL


Fueredoriku (approved):
Very nice feature! 🥳

Tested and code looks good 👍

PR #806, opened 2024-08-05
Fix issue where fire trails would not ignite players

### LINKED ISSUE #165: Add fire to bullet trails and mesh projectiles

Fueredoriku:


toberge (approved):
LGTM thank you for fix

PR #807, opened 2024-08-05
Editor-only graybox magma vents

Fueredoriku:


toberge (approved):
LGTM let's ship this /s

PR #808, opened 2024-08-05
Use lower volume for shotgun fire

toberge:


Fueredoriku (approved):
It still got the "oomph" but no longer cause ear drum explosions, very nice 👍

PR #814, opened 2024-08-17
Make lobby select match other menus w/info box

toberge:
This has bugged me for a while so I needed to fix it 😤


ThomasJGabrielsen (approved):
A small but noticeable improvement 👍 
bilde

PR #815, opened 2024-08-17
Move render texture script so builds work

toberge:
Had to be put in the Editor folder in order to not mess up builds. Also tested a mac build so there's a new file here.


Fueredoriku (approved):
Strange, but as long as it works...? 🤔

PR #816, opened 2024-08-17
Refactor track layers and update music

### LINKED ISSUE #812: Volume starts out lower than it is set?
Returning to main menu or going into options menu somehow fixes it

toberge:
~~⚠️ may have messed up the volume compared to the main menu and tutorial music, let me know if it feels off~~
no, that's the settings issue. fixed.

⚠️ something seems off with the fade-out + switching, the volume seems to start out lower? there seems to be an issue with looping as well 🤔 

# them changes

- [x] Update all the musics
- [x] Add separate brass layers
- [x] Refactor layer code
- ~~Only distort brass layer when hitting instrumentbot?~~ postponed as this would require individual variables for each layer, and then cause issues with backup layer -_-
- [x] Fix settings reloading every time you enter the main menu (static isLoaded bool)
- [x] Prevent volume from starting out at default levels every time you boot up the game (wait one frame problem)
- [x] Prevent resolution and display mode from being set to their current values again when applying settings (caused things to look a lil weird, at least for display mode)


Fueredoriku (approved):
The new versions of the songs are really nice! 😎 

The code also looks fine, but I noticed that my audio settings keep resetting to the deafult (all at max volume) even when I have settingsdata. When the game boots, the audio is as saved in the settingsdata previously, but as soon as I enter the settings menu, it resets to all sliders on max and the audio noticably immediately changes to max too. 

I would say this PR is good to go if this is fixed, so I'll leave an approve for you to use at your discretion. (pls fix first tho 🙏)

PR #818, opened 2024-08-27
Remove superflous is server check when yielding

### LINKED ISSUE #817: Yielding doesn't work when networked
Seems to be an issue where the check at the end of the yield countdown doesn't pass

Fueredoriku:


toberge (approved):
Tested, works well!

PR #819, opened 2024-09-08
Add WIP version of new Grand Canyon map

toberge:
Terrain by cristopher

- [x] Add WIP version of new Grand Canyon
- [x] Make it not appear in builds (separate list of maps in levelselectmanager)


Fueredoriku (approved):
Looks good! 👍

PR #820, opened 2024-09-08
Rebindable keys

### LINKED ISSUE #824: Keybindings

Fueredoriku:
- Add menu of unlockable customizations and extras (currently unaccesiable, to be finished in a later PR)
- Add rebindable keys to options menu (Movement + actions like crouch, jump, fire, aim)

Current rebindable keys are ONLY for keyboards, can be expanded upon in the future, but not in this PR 🤠

Note: Keys are not persistent when playing in the editor, only in the built version!


toberge (changes requested):
Seems to work for setting new binds and playing a round, but:

- The changes don't seem to persist
- The reset button is tiny
- We uh... shouldn't be able to do assign several actions/directions to the same key
- Longer key names become very tiny
- This of course doesn't affect the keybind loading screen tip (fine)
- And it doesn't use the custom keybind assets either (fine, haven't made more yet)

image

toberge (changes requested):
- Bidding movement keys don't apply to weapon construction, which is *very* jarring (and we'd likely fare much better if we actually overrode menu movement with the same keys here)
- Rebinding fire and jump keys doesn't change these keybinds in auctions

toberge (approved):
Seems to work great for the most part now!

I did find a few wrinkles, but these can be resolved later:
- Keybinds *are* persisted in the editor, you just need to trigger some refresh that happens by reloading the menu or trying to bind a key
- The pause menu should probably use the same binds as the in-game movement?

Fueredoriku:
Fixed the binding display refresh, but the pause menu stuff will be left for the poor sould who decides to complete #138

PR #821, opened 2024-09-16
Crosshair with modifiable radius and inverse colors

Fueredoriku:
- Crosshair with gunstats modifiable radius for the outline, and separate punch animation for only the cross.
- The crosshair now also has a black outline so it always has a nice contrast.
- Shotgun has a unique non-circular outline
- Hat-a-pult has a unique hat shaped outline
- Additional hitmarker when you deal damage (which is not always only bullet hit, this way we can differentiate between hits and fire procs)

### DIFF
--- a/Assets/Prefabs/GunParts/BulletBarrel.prefab
+++ b/Assets/Prefabs/GunParts/BulletBarrel.prefab
@@ -662,9 +662,14 @@ MonoBehaviour:
     addition: 0
     multiplier: 0
     exponential: 2
+  - name: CrosshairRadius
+    addition: -0.8
+    multiplier: 0
+    exponential: 1
   outputs:
   - {fileID: 5765902433490691002}
   midpoint: {fileID: 5139843727155892186}
+  crossHairMode: 0
   attachmentPoints:
   - {fileID: 5765902433490691002}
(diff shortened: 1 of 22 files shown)


toberge (changes requested):
In theory I very much agree with these changes, and I like the new hit animation and the overall size of the crosshair. However, there are some serious issues here:

The bullet barrel shouldn't have a circle. It doesn't make sense if you want to distinguish it from e.g. the laser barrel, where the shots have an actual radius. The radius for the laser barrel's crosshair actually matches very well with its shots.
image

Additionally, I find the inverted color look to be quite distracting. Most of the colors we have should not conflict that much with pure white, and the inversion doesn't even work when it is overlaid on top of explosions, so I'd suggest just disabling it for now, as the current version just ends up showing either white or soft blue most of the time.


Then there's the black border, which gets so few pixels that it looks kinda fuzzy. The shotgun crosshair has the border only around some of its edges. I'd say the crosshair shape looks fine, but if we are to have a border, it should at least be consistent.
image


As for the hat barrel, I got this suggestion from somebody, to emphasize its verticality:
image
Not sure if I'm entirely on board with that. Might just as well make a hat shape or sth 🤠

toberge (approved):
LVGTM

PR #823, opened 2024-09-19
Fix main menu freeze, resolves #822

### LINKED ISSUE #822: Not skipping intro cinematic locks players out of interacting with the menu

Fueredoriku:


Assets/Scripts/Control&Input/PlayerInputManagerController.cs
```diff
@@ -86,8 +86,9 @@ private void OnPlayerJoined(PlayerInput playerInput)
         var inputManager = playerInput.GetComponent<InputManager>();
         inputManager.PlayerCamera.enabled = false;
         LocalPlayerInputs.Add(inputManager);
+        if (playerInput.currentControlScheme == "MouseAndKeyboard")
+           inputManager.SetMouseAndKeyboard();
```
toberge, line 90: Couldn't we just put this check into the Start method for InputManager, since it sits on the playerinput and already accesses it, and thus avoid exposing the m&k bool?

Assets/Scripts/Control&Input/InputManager.cs
```diff
@@ -95,6 +95,11 @@ void OnDestroy()
         RemoveListeners();
     }
 
+    public void SetMouseAndKeyboard()
```
toberge, line 98: And besides, I'm not a fan of having setter-like methods without an actual input parameter. If we keep this method (see my other comment), perhaps rename it to `SetAsMouseAndKeyboard` so that its intent is a little more clear.

toberge (approved):
Looks sensible, I just have an implementation suggestion

Assets/Scripts/Control&Input/PlayerInputManagerController.cs
Fueredoriku: Nope, unfortunately due to the script execution order, this will return false as the add player delegate is invoked before input manager start is able to do anything, which is why I do it right before the line below this comment

Assets/Scripts/Control&Input/InputManager.cs
Fueredoriku: Yeah same, it is unfortunate that the state has to be updated by an outside class in this case

toberge: This bit isn't that important so I'll just resolve this comment ¯\_(ツ)_/¯

PR #827, opened 2024-09-25
Add sfx for walking and soda spray

Fueredoriku:
- Add walking sounds that are synced to walking animations and changes based on ground type
- Add spray noise to soda cans


Assets/Scripts/Interactables/SodaCan.cs
```diff
@@ -81,12 +87,16 @@ private void StartFlying()
...
+        audioSource.loop = true;
+        sprayNoise.PlayExclusively(audioSource);
     }
 
     private void StopFlying()
     {
         state = SodaCanState.Emptied;
         sprayEffect.SetBool("IsSpraying", false);
+        audioSource.loop = false;
+        audioSource.Stop();
```
toberge, line 99: Unsure if you need of these lines

toberge (approved):
The walking sounds work really well (once I turned down the music volume, mine was at max)

The soda spray sound could be more _continuous_, atm it is quite distracting compared to the more consistent noise one would expect

Assets/Scripts/Interactables/SodaCan.cs
Fueredoriku: These lines are here because both the explosion and the spray noise use the same audio source, so I need to ensure they dont overlap

toberge: Aha

PR #828, opened 2024-09-26
Improved pan extension

Fueredoriku:
- Pan now blocks explosion damage if aimed correctly
panSplosion
- Pan skating speed increased a lot
- Add pan-shot! popup text when blocking damage

panShot

### DIFF
--- a/Assets/Scripts/Augment/AugmentImplementations/Pan.cs
+++ b/Assets/Scripts/Augment/AugmentImplementations/Pan.cs
@@ -1,5 +1,6 @@
 using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 using UnityEngine;
 using UnityEngine.InputSystem;
 
@@ -8,11 +9,15 @@ public class Pan : GunExtension
     [SerializeField]
     private MeshCollider hitboxCollider;
     [SerializeField]
+    private GameObject explosionColliders;
+    [SerializeField]
     private Transform panModel;
     private PlayerMovement playerMovement;
-    private const float skateJumpForce = 5f;
+    private const float skateJumpForce = 11f;
+    private const float skateMoveForce = 2f;
     private GunController gunController;
     private const int hitBoxLayer = 3;
+    private const int aiExplosionLayer = 15;
     void Start()
     {
         gunController = transform.parent.GetComponent<GunController>();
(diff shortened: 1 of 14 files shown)


toberge (changes requested):
Nice that it finally blocks explosions 🎉 (though sometimes not quite)

However, currently it seems that bullets are blocked by the huge explosion catchers as well, which makes it impossible to hit someone from the front 🙃 
image

Assets/Scripts/Augment/AugmentImplementations/Pan.cs
```diff
@@ -22,30 +26,48 @@ void Start()
...
+            .ToList().ForEach(box =>  box.health = health);
+        if (playerMovement)
+            playerMovement.OnMove += TryPanSkateBoost;
     }
 
     private void TryTrickJump(InputAction.CallbackContext ctx)
     {
         if (playerMovement)
-            if (playerMovement.StateIsAir)
+            if (playerMovement.StateIsAir && playerMovement.IsCrouching)
```
toberge: at this point the `if` pyramid looks a little odd

Assets/Scripts/Augment/AugmentImplementations/PanHitbox.cs
```diff
@@ -0,0 +1,20 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class PanHitbox : HitboxController
+{
+    [SerializeField]
+    private HealthController panHealth;
```
toberge, line 8: Superfluous. The superclass already has a public health member.

Fueredoriku: They have two different purposes, the panhealth is for the panshotemitter, not for tracking actual damage!

Assets/Scripts/Augment/AugmentImplementations/Pan.cs
Fueredoriku: Hehehe, I concede that the pyramid should go, I will miss it tho 🫡

Assets/Scripts/Augment/AugmentImplementations/PanHitbox.cs
toberge: Oh okay

toberge (approved):
Works great now!

PR #829, opened 2024-09-28
Fix decals showing on tether plug

toberge:
Added a new tag for objects that should not have decals on them.

### DIFF
--- a/Assets/Scripts/Augment/BulletModifiers/DecalModifier.cs
+++ b/Assets/Scripts/Augment/BulletModifiers/DecalModifier.cs
@@ -23,12 +23,11 @@ public class DecalModifier : MonoBehaviour, ProjectileModifier
 
     [Range(0f, 180f)][SerializeField] private float angleVariation = 180f;
 
-    private const int allGunsAndPlayersMask = (1 << 3) | (1 << 8) | (1 << 9) | (1 << 10) | (1 << 11) | (1 << 12) | (1 << 13) | (1 << 14) | (1 << 15);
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Augment/BulletModifiers/DecalModifier.cs
```diff
@@ -57,13 +57,13 @@ public void Detach(ProjectileController projectile)
...
+        // Avoid placing decals on players or their guns, as that leads to the heebie-jeebies
         if ((target.collider.gameObject.layer & allGunsAndPlayersMask) > 0)
             return;
+        // Avoid decals on other objects that are marked as unsuitable for placing bullet holes on
+        if (target.collider.gameObject.CompareTag("NoDecal"))
+            return;
+        // And player hitboxes
         if (target.collider.TryGetComponent<HitboxController>(out var hitbox))
             if (hitbox.health && hitbox.health.TryGetComponent<PlayerManager>(out var _))
                 return;
```
Fueredoriku: Given we now check for the hitbox layer, maybe this line of code is superfluous?

Fueredoriku (approved):
I think most of the instances of "We don't want decals on this" tend to overlap with things that can be put in the hitbox layer, including soda and tetherplug. This might not always be the case for future "no decals pls"-objects though, so I'll give this the 👍

Assets/Scripts/Augment/BulletModifiers/DecalModifier.cs
toberge: Not necessarily. Decals should appear on the instrument bots, and they have hitboxes w/o having players tied to them... however, player hitboxes will already be excluded by the layer check so... yeah, we _should_ be able to get rid of this check? 🤔

PR #831, opened 2024-09-28
(Re)implement critical hits

### LINKED ISSUE #647: Headshots
Critical damage from headshot

toberge:
- [x] Crits increase damage
- [x] Working crit mult stat
- [x] Crits make dmg numbers a different color
- [x] Perhaps color grade dmg numbers now?

### DIFF
--- a/Assets/Prefabs/GunParts/CoilBarrel.prefab
+++ b/Assets/Prefabs/GunParts/CoilBarrel.prefab
@@ -414,6 +414,7 @@ MonoBehaviour:
   outputs:
   - {fileID: 5765902433490691002}
   midpoint: {fileID: 2658316667072905057}
+  crossHairMode: 0
   attachmentPoints:
   - {fileID: 5765902433490691002}
   muzzleFlash: {fileID: 261841362781268065}
--- a/Assets/Scripts/UI/PlayerHUDController.cs
+++ b/Assets/Scripts/UI/PlayerHUDController.cs
@@ -335,28 +335,38 @@ public class PlayerHUDController : MonoBehaviour
         if (LeanTween.isTweening(hitTween))
         {
             LeanTween.cancel(hitTween);
-            SetCrossScale(crosshairCrossScale);
+            SetCrosshairScale(crosshairCrossScale);
         }
-        hitTween = LeanTween.value(crosshair.gameObject, SetCrossScale, crosshairCrossScale, 2.5f, 0.4f).setEasePunch().id;
+        hitTween = LeanTween.value(crosshair.gameObject, SetCrosshairScale, crosshairCrossScale, 2.5f, 0.4f).setEasePunch().id;
     }
 
     // TODO better name for this method?
-    public void DamageAnimation()
+    public void DamageAnimation(DamageInfo damage)
     {
         if (LeanTween.isTweening(hitMarkTween))
         {
             LeanTween.cancel(hitMarkTween);
-            SetHitScale(0f);
+            SetHitmarkScale(0f);
+            SetCrosshairCrit(false);
         }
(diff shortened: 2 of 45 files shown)


Assets/Prefabs/GunParts/CoilBarrel.prefab
```diff
@@ -411,9 +411,14 @@ MonoBehaviour:
     addition: 1.2
     multiplier: 0
     exponential: 1
+  - name: CriticalMultiplier
+    addition: 0.3
+    multiplier: 0
+    exponential: 1
```
Fueredoriku: I think crits are best when they are a directly because the player took a risk to aim at a small area in order to get more reward (damage). I think the random nature of the coil makes it a bad candidate for headshot crits

Fueredoriku (commented):
Really nice to finally have crits as a good risk/reward system! 👌

I have some points I thinks we should address before merge though:
- We should have somewhere in the game inform players that headshots exists, maybe in the tutorial and/or loadingscreen (maybe tip users that some barrels do not grant headshot crits)
- The crit color is very similar in color to the cratertown background. Maybe do a gradient animation similar to the one on secret names in weapon select? (but maybe only use green and orange instead of the entire HSV rainbow lol)
- I think the crits need to have more feedback. They should probably have their own SFX (or at least a unique recognizable pitch), and we might also want to color the hitmarker in the crosshair red when critical damage is served

Assets/Prefabs/GunParts/CoilBarrel.prefab
Fueredoriku: I also think the Telescope could "grant" barrels that don't normaly have headshot crits the ability, like the coil, to do so

toberge: Rebalanced accordingly.

toberge:
> We should have somewhere in the game inform players that headshots exists, maybe in the tutorial and/or loadingscreen (maybe tip users that some barrels do not grant headshot crits)

Added a loading screen tip. No crit area on the dummies in training mode, so I plonked down some instrument bots to at least give the players _something_

> The crit color is very similar in color to the cratertown background. Maybe do a gradient animation similar to the one on secret names in weapon select? (but maybe only use green and orange instead of the entire HSV rainbow lol)

Turned the text color green.

> I think the crits need to have more feedback. They should probably have their own SFX (or at least a unique recognizable pitch), and we might also want to color the hitmarker in the crosshair red when critical damage is served

Crits now turn the crosshair green, and I made a pitch-shifted variant of the hit sounds for this. Ideally we'd also want different hit sounds for you vs other players, so we'll need to refurbish them at a later point -_-

Assets/Scripts/UI/PlayerHUDController.cs
```diff
@@ -335,34 +335,45 @@ public void HitAnimation(HitboxController other, ref ProjectileState state)
...
+            .setOnComplete(() => SetCrosshairCrit(false)).id;
+    }
+
+    private void SetCrosshairCrit(bool isCritical)
+    {
+        crosshairMaterial.SetInteger("_IsCritical", isCritical ? 1 : 0);
     }
 
-    private void SetCrossScale(float scale)
+    private void SetCrossharScale(float scale)
```
Fueredoriku: Missing an "i" in crosshair

Fueredoriku (approved):
Very nice stuff, I also love that we can distort the tutorial theme now 😅

I only have a small nitpick in a comment below, but other than that this seems great 👍

Assets/Scripts/UI/PlayerHUDController.cs
toberge: har har

PR #833, opened 2024-09-29
Add voice lines

toberge:
- [x] Add them voices
- [x] Place voice source on... camera? Some other node than base node, at least
- [x] Add echo effect to fall clips
- [x] Fix the problem with voice clips hanging? or is that gonna be solved by placing source elsewhere?
- [x] Credit voice actors


Fueredoriku (approved):
Very nice to have this!! I love the additional wackiness this feature brings 😁

PR #834, opened 2024-09-29
Mute soda can spray in splitscreen

toberge:


Fueredoriku (approved):
Probably a good idea... 😅

PR #835, opened 2024-09-30
Fix sound tag for proper auction scene walking SFX

Fueredoriku:
Because sand footsteps sounded a bit wrong...


toberge (approved):
Agreed, but why not wood? 🤔

Fueredoriku:
Sounded less... right? when I tested. I also like the idea that they used metal to imitate wood for the floor

PR #837, opened 2024-09-30
Fix type mismatch when setting crosshair critical

toberge:
Shader booleans are floats smh


Fueredoriku (approved):
Told ya! I have fallen to the same trap previously myself 😅

PR #840, opened 2024-10-01
Prevent nullreference in soda can sounds + misc other exceptionish fixes

toberge:
What could happen here, with the soda can hit sounds using the explosion's audio source:

- Soda can explodes
- Explosion destroys itself
- Hitbox is shot/damaged again
- No audiosource!

Also fixes a bunch of semi-related issues.


Fueredoriku (approved):
Seems good, code looks fine and manual testing didn't indicate anything wrong 👍

PR #842, opened 2024-10-01
Add sfx for bodies indicating they have no more ammo left

Fueredoriku:
Some nice extra feedback and a cue for players to understand that they have no ammo left.
The sfx is dependant on body, usually tied to firing mode and only where it makes sense to have this (not tether)

Also fixed a couple of missing desubscriptions for body delegates


toberge (approved):
probably fine, though ideally should not be implemented a bazillion times

Fueredoriku:
Will be addressed in #845

PR #844, opened 2024-10-02
Fix eventlog not being enabled on non-host clients

### LINKED ISSUE #841: No killfeed for non-host players
It seems only the host and local players on host is able to see the killfeed

Fueredoriku:
Closes #841


toberge (approved):
A little sus but if it work it do be working

PR #847, opened 2024-10-02
Fix broken revolver reload when spamming barrel with chargetime

### LINKED ISSUE #848: Revolver reload animation bug with coil (spam fire to reproduce)

Fueredoriku:
You can still initate the reload animation while firing the last time (looks kinda funny, so I think it's ok to leave in for now), but it will no longer mess up the reload animation/weapon!


toberge (approved):
LGTM, hope we can fix the lazur barrel's fire end invocation soon

PR #850, opened 2024-10-03
Use pragmatic synchronization for less jitter

toberge:
huehuehue


Fueredoriku (approved):
A lot better over bad connections, great stuff 👍

PR #860, opened 2024-10-06
Improve DDR body UI

toberge:
- Hollow target arrow
- Animation on move input
- Animation when succeeding

ddrgui


Assets/Scenes/Menu.unity
```diff
@@ -38,7 +38,7 @@ RenderSettings:
   m_ReflectionIntensity: 1
   m_CustomReflection: {fileID: 0}
   m_Sun: {fileID: 785370674}
-  m_IndirectSpecularColor: {r: 0.0624169, g: 0.21926133, b: 0.4947306, a: 1}
+  m_IndirectSpecularColor: {r: 0.054284044, g: 0.17437917, b: 0.38410002, a: 1}
```
Fueredoriku: 🤪

Fueredoriku (approved):
Very nice! 🕺

Could you also increment the version number? 🙏

Assets/Scenes/Menu.unity
toberge: ugh, fiiiiiine (though this shouldn't need to be reverted every single time it happens smh)

PR #861, opened 2024-10-06
Add thinner crosshairs + Crosshair size setting

### LINKED ISSUE #859: Crosshair size slider

Fueredoriku:
- Add thinner crosshairs
- Add crosshair size setting top options
- Fix crosshair aliasing with smoothstepped black border


toberge (changes requested):
The normal crosshair is appropriately thin now (at size 1.0), nice!

However,
- The shotgun and hat crosshairs seem a little off compared to the rest. The cross doesn't feel nearly as thin. I do know that these barrels don't work the same when aiming over long distances, but having two different thicknesses on the cross in the middle feels... very odd
- When no settings file is present, the size defaults to 0 💀

Assets/Scripts/SettingsDataManager.cs
```diff
@@ -64,6 +76,9 @@ public SettingsData(SettingsDataStruct settingsData)
         SfxVolume = settingsData.SfxVolume;
         QualityPresetIndex = settingsData.QualityPresetIndex;
         DisplayModeIndex = settingsData.DisplayModeIndex;
+        // Settings added after this needs to be backwards compatible with players who saved old setttings
+        // This can be achieved by checking for larger than 0
+        CrosshairSize = settingsData.CrosshairSize > 0 ? settingsData.CrosshairSize : 1f;
```
toberge: You um... forgot to use the default constant here :)))))

Assets/Shaders/CrossHairShotgun.hlsl
```diff
@@ -37,5 +37,6 @@ float shotgun_float(float2 pos)
 
 void CrossHairShotgun_float(float2 UV, float crossSize, float circleRadius, float hitMarkerRadius, out float Distance)
 {
-	Distance = min(min(shotgun_float(UV * circleRadius), crossSDFShotgun_float(UV * crossSize, 1.)), sdRoundedXShotgun_float(UV, hitMarkerRadius, 0.1));
+    UV *= 0.9;
```
toberge, line 40: Huh, so you just made the whole thing smaller?
What I wanted to point out was the jarring difference in thinness between this cross
image
and this cross
image
which has become even greater now that they differ this much in length.
As for the hat, it's still a lot thicker than the rest:
image
(which feels fine for the hat, but it seems more ideal to have the crosses be of the same shape)
And with the telescope extension, it has uhhh... the same issue as before this PR
image

toberge (approved):
The default crosshair size works now 👍 

I leave an approve here, but I still have some concerns, ref. comments

Assets/Scripts/SettingsDataManager.cs
Fueredoriku: Fixed 👍

Assets/Shaders/CrossHairShotgun.hlsl
Fueredoriku: It was both, the crosshair became thinner, shotgun specifically was additionaly shrunk.
The observed thickness difference stems from the center being rounded towards the arms of the cross, thus seeming thicker than they actually are for crosshairs with shorter "hands"

Hat now has an even smaller non-reactive cross instead, with the punch bein applied to the hatSDF instead

PR #862, opened 2024-10-11
Stop AIs from shooting at dead players + more

### LINKED ISSUE #533: Make AIs only prioritize each other when no real players

toberge:
Also closes #533

### DIFF
--- a/Assets/Scripts/Gamestate/AIManager.cs
+++ b/Assets/Scripts/Gamestate/AIManager.cs
@@ -48,6 +48,8 @@ public class AIManager : PlayerManager
     private delegate void NavMeshEvent();
     private NavMeshEvent onLinkStart;
     private NavMeshEvent onLinkEnd;
+    private AmmoBoxCollector ammoBoxCollector;
+    private Coroutine airDisablingRoutine;
 
     private void Start()
     {
@@ -58,6 +60,7 @@ public class AIManager : PlayerManager
         colliderBox = GetComponent<Collider>();
         colliderBox.isTrigger = true;
         aiMovement = GetComponent<AIMovement>();
+        ammoBoxCollector = GetComponent<AmmoBoxCollector>();
         healthController = GetComponent<HealthController>();
         agent.autoTraverseOffMeshLink = false;
         onLinkStart += AnimateJump;
@@ -135,7 +138,7 @@ public class AIManager : PlayerManager
         playerIK.LeftHandIKTarget = gunController.LeftHandTarget;
         if (gunController.RightHandTarget)
             playerIK.RightHandIKTarget = gunController.RightHandTarget;
-        GetComponent<AmmoBoxCollector>().CheckForAmmoBoxBodyAgain();
+        ammoBoxCollector.CheckForAmmoBoxBodyAgain();
     }
 
     private bool IsDisabledItem(Item item)
@@ -154,22 +157,33 @@ public class AIManager : PlayerManager
         }
         if (info.damageType != DamageType.Explosion)
             return;
-        StartCoroutine(WaitAndToggleAgent());
+
+        if (airDisablingRoutine != null)
+            StopCoroutine(airDisablingRoutine);
+        airDisablingRoutine = StartCoroutine(WaitAndToggleAgent());
     }
 
-    public IEnumerator WaitAndToggleAgent()
+    private void DisableAgent()
     {
         agent.enabled = false;
         body.isKinematic = false;
         colliderBox.isTrigger = false;
+    }
 
-        yield return new WaitForSeconds(0.25f);
-        while (aiMovement.StateIsAir)
-            yield return null;
-
+    private void EnableAgent()
+    {
         body.isKinematic = true;
         agent.enabled = true;
         colliderBox.isTrigger = true;
+    }
+
+    public IEnumerator WaitAndToggleAgent()
+    {
+        DisableAgent();
+        yield return new WaitForSeconds(0.25f);
+        while (aiMovement.StateIsAir)
+            yield return null;
+        EnableAgent();
     }
 
     protected override void OnDeath(HealthController healthController, float damage, DamageInfo info)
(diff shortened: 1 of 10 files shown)


Assets/Scripts/Gamestate/AIManager.cs
```diff
@@ -195,32 +210,20 @@ private IEnumerator LookForTargets()
...
+        var closestDistance = ignoreAwareRadius;
 
-            closestPlayer = player.AiTarget.transform;
-            closestDistance = hitDistance;
-            DestinationTarget = closestPlayer;
-            ShootingTarget = player.AiAimSpot;
+        // TODO the ammo box stuff!
+        var hasAmmoBoxBody = ammoBoxCollector.CanReload;
+        var isOutOfAmmo = gunController && gunController.stats.Ammo < 1;
+        if (hasAmmoBoxBody && isOutOfAmmo)
```
Fueredoriku: This should also be called when the AI is in close proximity mode! If the AI depletes its ammo and is in close range it'll soft lock itself not being able to shoot and unwilling to find new ammo

Assets/Scripts/Gamestate/AIManager.cs
```diff
@@ -237,10 +240,23 @@ private void FindPlayers()
             aiMovement.Target = ShootingTarget;
             var isStrafeDistance = closestDistance < 10f;
             aiMovement.enabled = isStrafeDistance;
+
+            var closestExplodingBarrel = ExplodingBarrel.GetViableExplodingBarrel(ShootingTarget.position);
+            var isBarrelSafeToShoot = closestExplodingBarrel
+                && Vector3.Distance(closestExplodingBarrel.transform.position, transform.position) > closestExplodingBarrel.Radius + 1;
+            if (isBarrelSafeToShoot)
+            {
+                Debug.Log($"Found exploding barrel at {closestExplodingBarrel.transform.position} >:)", closestExplodingBarrel);
```
Fueredoriku: remove debug log?  🙏

Fueredoriku (changes requested):
Great work, the AIs become suprisingly more fun to play against when they can use seeker and tether too!
 
I have identified three issues that need fixing:
- Ammoboxes ignored if AI has entered a close range state
- Tether may never be replugged
(I suggest we fix the tether problem by forcing the AIs to always aim straight down when planting the plug)
tethersoftlock



- Also it seems that if the player is gone, the AIs may sometimes choose to stand still at some place and shoot the remaining targets from that single spot only. To replicate this problem, make one or more of the AIs target you, die, then spectate and there is a good chance that they may get stuck standing in a specific place. See one of the comments I made for an example.

The rest looks good here, so if we fix these issues this should be good to go!

Assets/Scripts/Gamestate/AIManager.cs
```diff
@@ -260,6 +276,55 @@ private void FindPlayers()
...
+            if (areOnlyAiPlayersLeft)
+            {
+                if (hitDistance > closestDistance)
+                    continue;
+
+                closestDistance = hitDistance;
+                closestPlayer = player.AiTarget.transform;
+                ShootingTarget = player.AiAimSpot;
+                continue;
+            }
```
Fueredoriku, line 339: There seems to be a problem with something inside the areOnlyAiPlayersLeft, making it so the AIs can choose to stand still firing at eachother from places they can't hit eachother, thus softlocking the game

Here is an example spot from The Pit
image

Assets/Scripts/Gamestate/AIManager.cs
toberge: Removed 😔

Assets/Scripts/Gamestate/AIManager.cs
toberge: Reworked so it happens when they are in close proximity, _without_ messing up their trajectory if they are in the air

Assets/Scripts/Gamestate/AIManager.cs
toberge: Fixed. Removed the ShootingTarget and continue; lines, instead setting DestinationTarget.

toberge:
> Great work, the AIs become suprisingly more fun to play against when they can use seeker and tether too!
> 
> I have identified three issues that need fixing:
> 
> * Ammoboxes ignored if AI has entered a close range state

Fixed, see other comment

>  * Tether may never be replugged
>     (I suggest we fix the tether problem by forcing the AIs to always aim straight down when planting the plug)

Fixed as suggested


> * Also it seems that if the player is gone, the AIs may sometimes choose to stand still at some place and shoot the remaining targets from that single spot only. To replicate this problem, make one or more of the AIs target you, die, then spectate and there is a good chance that they may get stuck standing in a specific place. See one of the comments I made for an example.

Fixed. They don't instantly set ShootingTarget here anymore, instead DestinationTarget so that they walk to their enemy first.

Assets/Scripts/Gamestate/AIManager.cs
```diff
@@ -186,61 +227,78 @@ private IEnumerator LookForTargets()
...
+                var closestExplodingBarrel = ExplodingBarrel.GetViableExplodingBarrel(ShootingTarget.position);
+                var isBarrelSafeToShoot = closestExplodingBarrel
+                    && Vector3.Distance(closestExplodingBarrel.transform.position, transform.position) > closestExplodingBarrel.Radius + 1;
+                if (isBarrelSafeToShoot)
+                {
+                    // TODO respond to the barrel actually exploding?
+                    // TODO the shooting target going missing seems to be the cause for 
+                    ShootingTarget = closestExplodingBarrel.transform;
+                    // TODO unsub?
+                    closestExplodingBarrel.GetComponent<HealthController>().onDeath += HandleExplodedBarrel;
```
Fueredoriku, line 289: Maybe unsubscribe at the HandleExplodedBarrel function itself? Might be nice in order to minimize memory leakage

Fueredoriku (approved):
Great work!! 👍

The AIs seem a lot more... sane now 😅

I left a small comment but other than that I have no more feedback, so I'll leave an approve for you to use at your discretion

Assets/Scripts/Gamestate/AIManager.cs
Fueredoriku: Nice!

Assets/Scripts/Gamestate/AIManager.cs
toberge: Thought I did that already but must have forgor 💀

toberge: Nope, I already did every single thing that the TODO comments around here were talking about. Removed the superfluous comments now 🙃

PR #869, opened 2024-10-23
wip fix ai air stuff

toberge:


toberge:
@Fueredoriku does this make a smidge of sense?

Fueredoriku:
I think it's a good idea to check if the AI is grounded and use that to force the AI to only setDestination when grounded. Using the stepping(edging lol)-check is a good way to do that, as long as we make sure to modify the collider a bit. I think the stepping collider has to be a lot thinner to avoid still getting stuck on walls in the pit.

toberge:
The more I dig into this, the more it seems that the AI code is in dire need of being refactored into a state machine so we have an easier time working on it...

PR #870, opened 2024-10-26
Fix lazur barrel missing animation end + misc related fixes

toberge:


Fueredoriku (approved):
Nice fix, some of this also overlap with the zoom fix I'm working on, so definitely healthier code 👍

PR #875, opened 2024-11-15
Fix keybinding persistence

toberge:


Fueredoriku (approved):
Good fix, MASSIVE KeyMappingButton lint moment 👍

PR #876, opened 2024-11-15
Add look inversion settings

### LINKED ISSUE #874: Invert look option (mwahaha)
As requested in the Discord server

toberge:
Closes #874


Fueredoriku (approved):
Looks good and seem to work as expected! 👍

My only feedback is that I would consider refactoring `SettingsDataInstance` to something more descriptive than just `Data`, but I'm not picky so I'll leave an approve anyway 😎

toberge:
> My only feedback is that I would consider refactoring `SettingsDataInstance` to something more descriptive than just `Data`, but I'm not picky so I'll leave an approve anyway 😎

My problem with the original name was that it repeats the information that is already present in the class name, and in all cases (I think) the data instance thingy is accessed through the whole SettingsDataManager.SIngleton shebang. I'd like to resolve this more cleanly with read-only properties that access a _private_ instance of the SettingsData, but that'll have to happen in another PR.

PR #882, opened 2024-11-28
Make it possible to jump off of explosive barrels

toberge:
Added a tag for things that should not be tetherable, as the explosive barrel is the *only* exception so far where we can't just set a different layer.


Fueredoriku (changes requested):
image
Doesn't seem to stop the tether from placing lines? 🤔

Also I think we could name the tag something like "NotStatic", as the thing tether (and indeed some other things might want to avoid) is objects that are able to move

Fueredoriku (approved):
Good stuff 👍

PR #885, opened 2024-12-15
Add more weapon lore

toberge:


Fueredoriku (approved):
This is very much inline with my head cannon,  I'll probably add some more soon too 👍

PR #894, opened 2024-12-20
Brick barrel

### LINKED ISSUE #877: Brick barrel
Shoots bricks that inherit your momentum and deal more damage the faster they move.

Could have a speedometer mounted on the barrel

Suggested by Uwan368 on Steam

toberge:
- [x] Make it possible to pass rotation to mesh projectiles
  - doesn't function properly at the moment, but works well enough for erratic brick movement :sunglasses:  
- [x] Refactor VFXPositionFormatter to be a regular class and not a MonoBehaviour
- [x] Implement velocity inheritance
- [x] Implement velocity-by-damage
- [x] Balance damage graph
- [x] Make damage scale in explosion as well
- ~~Perhaps do explosion dmg as its own stat?~~
- [x] Model a brick
- [x] remember to make the brick red
- [x] Create brick explosion
   - [x] Brick splinters
   - [x] White cloud stuff
   - [x] Outline?
- [x] Model the barrel
   - [x] Speedometer
   - [x] Brick output (rectangular thingy)
   - [x] ????
- [x] Animate pistons
- [ ] Animate bricks???
- [x] Write speedometer shader
- [ ] Prettify speedometer
- [x] Make this work over network :/
- [x] Make AIs able to use it??? (determine if still or walking at least)

Closes #877 

velocity thresholds:
- 0: still
- 4: crouch
- 10: walk
- 11/12: leap
- 15: first bunnyleap
- 17: second bunnyleap
- 22+: falling velocity
- 25: top speed

image


Fueredoriku (approved):
Very fun augment! 👍
- Rendering layer mask of the mesh renderer should be foreground objects instead of light layer default to avoid decals projecting onto the augment
- The speedometer material should be unlit so players always can see the values clearly

Also make double sure that all VFXPosition buffers are released, it seems they are from what I can tell, but I'm scared of the potential memory leaks lol

I'll leave an approve you can use after the you fix the material/render mask stuff 😁👍

PR #902, opened 2024-12-21
Fix missing sky on Linux

toberge:
Turns out the start shader had a case difference, which breaks on Linux and not on Windows.


Fueredoriku (approved):
Cursed Windows

PR #903, opened 2024-12-22
Add weapon bobbing for walking and leaping

### LINKED ISSUE #892: Weapon bobbing
Weapon sway when walking

Fueredoriku:
May seem subtle, but is very satisfying!


toberge (approved):
feels good and does not seem to break other things

PR #909, opened 2024-12-23
Avoid rendering the entire scene twice each frame

### LINKED ISSUE #908: Avoid rendering two cameras 💀
So what we need to do is:
- [ ] Disable player camera when viewing panning camera
- [ ] Disable panning camera when:
  - [ ] joining game with just one player
  - [ ] in splitscreen but noone has died yet
- at start of round the panning camera should be enabled; disable it when player is set up!

which means
- by default enabled
- disable panning cam when any player has their setplayerinput method called
- enable panning cam when player has died basically (could be more granular here)

Should result in a whopping **100%** performance increase.

Might resove #804 actually.

toberge:
The overview camera has been enabled in the background and caused double rendering load *this entire time*. These changes should fix the problem, but some hiccups will still happen when playing in splitscreen and switching to the overview camera after one player is dead.

Closes #908

### DIFF
--- a/Assets/Scripts/Control&Input/OrbitCamera.cs
+++ b/Assets/Scripts/Control&Input/OrbitCamera.cs
@@ -95,8 +95,16 @@ public class OrbitCamera : MonoBehaviour
     {
         if (!cameraTransform || !input || MatchController.Singleton.IsShowingScoreboards)
             return;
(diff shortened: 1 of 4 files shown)


Fueredoriku (approved):
Well, better late than never 😅

PR #910, opened 2024-12-24
Add santa hats

toberge:
- [x] time limit to xmas?


Assets/Scripts/PlayerCosmetics.cs
```diff
@@ -1,4 +1,7 @@
+using Newtonsoft.Json.Linq;
```
toberge: bruh...

Fueredoriku: ooops, fixed

Fueredoriku (approved):
I might be biased, but I will approve 😅

PR #912, opened 2024-12-26
Split development info out to CONTRIBUTING.md

toberge:


toberge:
Yeah I thought I'd write some more here, but alas

Fueredoriku (approved):
Looks nice, maybe we should  also update the gifs and the change the video link to the early access trailer while we're at it. You don't have to for this PR though, so I'll leave an 👍 for you

PR #915, opened 2024-12-27
Fix fan hitbox being too low, fix particles spawning below fan

Fueredoriku:
Fixes the reported bug issue where someone got stuck underneath a fan
image


toberge (approved):
Yup, the hitboxes look much more reasonable now :+1:

PR #916, opened 2024-12-27
Fix detonate exceptions from async server calls with pooling

Fueredoriku:
A bug report seemed to indicate null pointer exceptions when detonating and toggling renderers. This can be caused by dynamites being pooled while recieving an asynchronous detonate call. To mitigate this, if guards have been introduced when toggling from the detonate call.

A try catch block has also been added, as to mitigate the consequences of similar issues in the future, as has been done for RPCFire in GunController for the same reasons.

### DIFF
--- a/Assets/Scripts/Augment/AugmentImplementations/StickyDynamite.cs
+++ b/Assets/Scripts/Augment/AugmentImplementations/StickyDynamite.cs
@@ -44,10 +44,18 @@ public class StickyDynamite : StuckObject
 
     public void Detonate(PlayerManager sourcePlayer)
     {
-        mesh.enabled = false;
-        explosiveBarrel.enabled = false;
-        smoke.enabled = false;
-        barrelSmoke.ToList().ForEach(smoke => smoke.enabled = false);
+        // Needs extra if-guards in case resetdynamite is called asynchronously by server
+        if (mesh)
+            mesh.enabled = false;
+        if (explosiveBarrel)
+            explosiveBarrel.enabled = false;
+        if(smoke)
+            smoke.enabled = false;
+        barrelSmoke.ToList().ForEach(smoke =>
+            {
+                if (smoke)
+                    smoke.enabled = false;
+            });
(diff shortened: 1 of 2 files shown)


Assets/Scripts/Augment/AugmentImplementations/StickyDynamite.cs
```diff
@@ -44,10 +44,18 @@ public void SetBarrel()
 
     public void Detonate(PlayerManager sourcePlayer)
     {
-        mesh.enabled = false;
-        explosiveBarrel.enabled = false;
-        smoke.enabled = false;
-        barrelSmoke.ToList().ForEach(smoke => smoke.enabled = false);
+        // Needs extra if-guards in case resetdynamite is called asynchronously by server
```
toberge, line 47: Dunno about the if guards here, might be that the try-catch block is enough as a safeguard for crashes. However, if we end up with a dynamite stick that is in a bad state for some reason... sure, keep it in

toberge (approved):
Probably avoids some of the issues, but that bug report might've been related to something that could happen if you detonate at the very start of a round perhaps?

toberge:
Actually, the reported bug wasn't found during online play at all AFAICT, it just looks that way since everything behaves kinda like it's online now. Thus, it might not have been an async issue at all...

Fueredoriku:
I am very sure this is due to timing errors as the RPC's on hosting clients that are fast (like if they game is built and not in editor) can be called immidiately, while non-hosting client Rpc's will end up in an execution queue and handled at the end of a frame (based on what I managed to find in the documentation). It might be wrong to write "asynchronous" in the threading sense, but I believe this is a timing issue between the pooling of sticky dynamites and the timing of the rpc detonate. Either way, I think the if guards are worth adding 😅

toberge:
Yeah, the guards and catch should alleviate both this and future issues, but we don't know if this was during online play (it likely wasn't), and this fix may hide the real cause for issues like this one (though more severe effects may be visible enough to be reported). I've left an approve here already, so feel free to slam the fix in 🫡

PR #917, opened 2024-12-27
Fix telescope sight persisting after death in spectator mode

Fueredoriku:
Fix for reported issue where spectator text would be scrambled on zooming in spectator mode if the player was using the telescope.


Assets/Scripts/Augment/AugmentImplementations/Telescope.cs
```diff
@@ -1,3 +1,4 @@
+using System;
```
toberge: Why's System imported?

toberge (approved):
LGTM just remove the import if it wasn't needed

Assets/Scripts/Augment/AugmentImplementations/Telescope.cs
Fueredoriku: Fixed

PR #918, opened 2024-12-27
Fix unintended visual yield countdown reset, resolves #852

### LINKED ISSUE #852: Yield sign resets countdown visually when new player enters

Fueredoriku:


toberge (commented):
This should solve _that_ yield problem, but also seems like it should have been handled by keeping a reference to the coroutine. Will have to test if this breaks when you wanna *stop* yielding

toberge (changes requested):
Seems to break when walking away from the yield area? As in, it seemed like the countdown still got to 0 (just not visually) even though I'd stepped out *and* the sign changed back?

Assets/Materials/CraterTown/transparentGlass.mat
```diff
@@ -24,10 +24,7 @@ Material:
   m_Shader: {fileID: -6465566751694194690, guid: 6dfb8dada865e684ab7b6b4c4131b184, type: 3}
   m_Parent: {fileID: 0}
   m_ModifiedSerializedProperties: 0
-  m_ValidKeywords:
-  - MAIN_LIGHT_CALCULATE_SHADOWS
-  - _MAIN_LIGHT_SHADOWS_CASCADE
-  - _SHADOWS_SOFT
+  m_ValidKeywords: []
```
toberge: oh whoops, should have made custom material for the vials on the brick barrel :skull: 
this is used for the glass in grand canyon so please create a new one here? ...or, I could do it myself I suppose :/

Assets/Scripts/Gamestate/AIManager.cs
```diff
@@ -71,6 +71,9 @@ private void Start()
         aiTargetCollider = Instantiate(aiTarget).GetComponent<AITarget>();
         aiTargetCollider.Owner = this;
         aiTargetCollider.transform.position = transform.position;
+        // Only look for FPS targets in non-auction settings
+        if (FindAnyObjectByType<AuctionDriver>())
```
toberge: Could be determined by checking for matchmanager as well (it is null when in auction)
...
or, hold on, AuctionDriver is a singleton too, so couldn't you just check that one for nullness?

toberge (approved):
Seems to work much better than before!

Assets/Scripts/Gamestate/AIManager.cs
Fueredoriku: Now checks singleton instead

PR #927, opened 2025-01-08
Fix loadingscreen transition flicker at start of animation

Fueredoriku:
Started noticing a very annoying flicker at the start of the loading screen transition animation.
This was due to the transition starting at the finished animation state the fram it was set active, now it always is enabled with the the correct first frame.


toberge (approved):
LGTM

PR #930, opened 2025-01-22
Add fancier border on loading screen transition

Fueredoriku:
Now it looks more like I originaly envisioned👌


toberge (approved):
laplacian enjoyer

PR #932, opened 2025-01-29
Add more textures to cowbots

Fueredoriku:
New fancy textures for cowbots! Includes much needed normal map for player-color material

rotatenormal


toberge (approved):
yolo

PR #935, opened 2025-02-02
Better visibility for players at a distance, and victory screen outline animation

### LINKED ISSUE #335: Highlight Players (Outline)
Outline colored after players color around player that is not adjusted for depth so we can see the player at least a s blob from any point on the map

Fueredoriku:
- Made players change LOD levels change at more reasonable ditstances
- Add inverse hull outline shader and material
- Made LOD3 players be transparent with inverse hull outline for better visibility
- Add jump fill outline shader and material
- Add jump fill outline to players seen from the arena camera (for spectating from far away + through geometry)
- Add jump fill outline animation to victory screen


Assets/Shaders/RFOShader.shader
toberge, line 1: perhaps some comment on top of this file with accreditation and license since this was from that article guy right?

Assets/Shaders/RFO.cs
toberge, line 1: also counts for parts of this file or?

Assets/Scripts/Scoreboard/Scoreboard.cs
```diff
@@ -229,4 +229,15 @@ private string FormatScore(int score)
...
     }
+
+    public void SetUIMask(bool isMasking)
+    {
+        // UI or default
+        var layer = isMasking ? 5 : 0;
+        gameObject.layer = layer;
+        foreach (Transform child in transform)
+        {
+            child.gameObject.layer = layer;
```
toberge, line 240: Not recursive?

toberge (approved):
Looks good for the most part, however

- Haven't inspected LOD stuff closely so only speaking about jump flood here, LOD seemed fine
- Victory outline animation has a jarring transition between thick black outline (very visible in some cases) and the extra thick white outline, which... kinda throw me off a little? Perhaps the swoosh to white should happen only when the camera has gone still (or just a little before), and perhaps we could avoid having the thick black outline displayed before it? (it cuts through the tower floor in crater town very noticeably)
- Feels like the player outlines on overview camera should be a different color than black? Perhaps white there as well? Looks kinda grimy :/ (my first thought here was to use the player colors but... this is already nice, that stuff can be attempted later)

Assets/Scripts/Scoreboard/Scoreboard.cs
Fueredoriku: Don't actually need to because the outline shader pass creates a mask that already covers the relevant parts of the screen 🤠

Assets/Shaders/RFOShader.shader
Fueredoriku: Added credits, license is a "do whatever you want", so I'll leave it at credits to article 👍

Assets/Shaders/RFO.cs
Fueredoriku: Yup

Fueredoriku:
Changed up the victory animation a bit based on your feedback and it looks a lot better now imo. Also added white outlines to players on the overview camera

PR #939, opened 2025-02-05
Add slightly better explosion screen shake

Fueredoriku:
Shake is more noticable and less dependant on viewing angle


toberge (commented):
Still don't really notice the shake 🤔

toberge (approved):
like, I can approve but I don't see it actually?

PR #941, opened 2025-02-05
Add materials to cratertown shacks, add wind particles!

Fueredoriku:
- Add textures to shacks
- Add wind particles to all maps
- Add wind particles to fans
- Add spraycans to main menu

### DIFF
--- a/Assets/Materials/ThePit/metalTextureGreenLight.mat
+++ b/Assets/Materials/ThePit/metalTextureGreenLight.mat
@@ -1,0 +1,164 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!114 &-832440985124310914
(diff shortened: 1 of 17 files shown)


toberge (approved):
omg the wind flow so nicely :o

only thing is that it looks a little too grey and blurry up close, perhaps we could make it transition its alpha by steps and not gradually?

but yeah it finally looks like the fans produce wind and not spiral helixes

re: the shacks, some of the color choices seem a little garish (the red in particular) but they do kinda work, and some of the wood materials there are perhaps a tad too similar to the sand... but it mostly works

though do note that I'm looking at the shacks with a blue light filter so they probably look fine :))))

Fueredoriku:
I'll put up an issue for discrete step particles 😎👍

PR #951, opened 2025-02-09
Fix decals being able to attach to various bullet hole decals

Fueredoriku:
Experienced an issue where an explosion mark would attach itself to a bullet hole decal, which would move around on the map when pooled. Solved by setting all three pooled bullet decals to hitbox layer.

Also reverted the cratertown glass dither material, because it doesn't look good in motion. We would need fractal benday dots dithering or something for that to work imo


Assets/Materials/CraterTown/DitherGlass.mat
toberge, line 1: Should use one of the transparentGlass materials instead so we avoid name confusion

toberge (approved):
Probably fine, not sure how to verify the decal fix

Assets/Materials/CraterTown/DitherGlass.mat
Fueredoriku: done 👍

PR #952, opened 2025-02-10
Fix bfl + revolver firing and reloading issues

Fueredoriku:
- Fix lazur not consuming ammo due to too late animation event trigger
- Fix lazur shooting an extra beam with revolver while reloading

### DIFF
--- a/Assets/LazurFiringAnimator.cs
+++ b/Assets/LazurFiringAnimator.cs
@@ -7,6 +7,8 @@ public class LazurFiringAnimator : AugmentAnimator
 
     public AnimationEvent OnChargeStart;
 
+    public bool IsDisabledInReload = false;
+
     public override void OnInitialize(GunStats stats) { }
 
     public override void OnReload(GunStats stats) { }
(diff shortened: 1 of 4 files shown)


Assets/LazurFiringAnimator.cs
```diff
@@ -28,6 +30,8 @@ public void EndFiring()
 
     public override void OnFire(GunStats stats)
     {
+        if (IsDiabledInReload && stats.Ammo <= 1)
```
toberge: This should actually be in all the firinganimators I think (at least for coil animator as well since it also is possible to start firing that one during reload)

toberge (approved):
Fiksen funker, og bør også brukes for andre våpendeler.

Ideelt sett burde vi dog få tilbake isFiring som jeg begynte med i tidligere forsøk på å fikse dette problemet, men det er betydelig mer kinkig.

Assets/LazurFiringAnimator.cs
Fueredoriku: It could, but I'm very worried about unforseen regressions on some of the implementations. I'll add it to the coil manually for now and I'll put up an refactor issue we can do after the update so we have the time properly test before this is released 👍

PR #958, opened 2025-02-28
Fix flying platform shadows

Fueredoriku:
The inner part of the shadow was missing due to solar material, fixed by adding an opaque plane inside the object


toberge (changes requested):
image

um... you didn't fix this in the *real* flying platforms 🙃

toberge (approved):
ye well whatever

PR #964, opened 2025-03-11
Synchronize yielding to fix jank

toberge:
explicit synchronization is the way to go

- should the flickering yield sign problem
- won't avoid problems where clients try to yield/stop yielding but aren't in sync (would take more refactoring)
- should also fix some problems with the yield countdown not restarting maybe


Fueredoriku (approved):
Code looks good, great to have this fixed!! 👍
We should probably test this a bit thoroughly together before we deploy this (but the behaviour can't be much worse than it already is lol)

PR #965, opened 2025-03-12
Slightly improved AI

### LINKED ISSUE #867: Fix AIs having perfect aim with the rubber sniper
Add AI aim accuracy gun stat

### LINKED ISSUE #962: Add more restictive AI navmesh on grand canyon
Stop AIs from walking on top of the map

Fueredoriku:
- Made AIs adjust their jumping animations based on the length of the jump
- Made AIs stop crouching after jumping
- Made AIs jump slightly up before jumping down to make down jumps a bit less jank
- Imporved navmesh in the Pit A LOT
- Add tighter map boundaries for players and AIs in grand canyon
- Add new loading screen tips and a new crime
- Fixed #867 

Also I think the reason AI agents freeze a bit and suddenly moves after jumping might be due to asynchronous update of their paths that happens when SetDestination() is called (Based on some threads I read on the forums like: https://discussions.unity.com/t/many-navmesh-agents-makes-causes-them-to-freeze-creates-delay-with-input/147200). This means the freeze timing become worse on low end machines. An alternative might be to force it to be calculated in the main thread by replacing almost all setDestination() with CalculatePath and caching the path, but that is way too much for me to do in this PR


Assets/Scripts/Gamestate/AIManager.cs
```diff
@@ -216,8 +224,11 @@ private void UpdateAimTarget(GunStats stats)
     {
         if (!ShootingTarget)
             return;
+        var badAimMultiplier = hasInaccurateAugment ? 4f : 1f;
+        if (hasInaccurateAugment)
+            Debug.Log("Rubber lol");
```
toberge: stray debug log 😔

Fueredoriku: Removed!

Assets/Scripts/Gamestate/AIManager.cs
```diff
@@ -371,17 +381,36 @@ private Transform FindClosestPlayer(out float closestDistance)
...
+                {
+                    animator.SetBool("Crouching", true);
+                    animator.SetTrigger("Leap");
+                }
+                StartCoroutine(AnimateJumpCurve(distance * 0.1f, offMeshLinkData.linkType));
+            }
+        }
+    }
+
+    private IEnumerator JumpCoolDown()
```
toberge: du kaller ikke denne fra noe sted? 🤔

toberge (approved):
yeah let's get this in!

Assets/Scripts/Gamestate/AIManager.cs
Fueredoriku: Eksperiment jeg glemte å fjerne

PR #966, opened 2025-03-13
Fix more menu jank

toberge:
death to minwidth

test at different aspect ratios

still haven't fixed the landing/lobby/gamemode buttons


Fueredoriku (approved):
Seems to be working as intended, not seeing the x-shifted gamemodes anymore at least 😅

PR #967, opened 2025-03-14
Add extra props to all maps and menu, add windmill and solarpanels

Fueredoriku:
I couldn't help myself 🤠

- Added animated windmill prop
- Added solar panel prop
- Added props to all maps
- Added props to main menu
- Added spikes to the pit 
- Made wind particles slightly more common


toberge (approved):
LGTM

PR #968, opened 2025-03-14
Add debug logs that may prevent a certain crash

toberge:
This may happen because logging increases the time it takes before the players are considered ready and input can be registered and funneled into Command calls. Or I may have gotten lucky while testing. Who knows.


ThomasJGabrielsen:
Pull request looks fine, but what is the crash and how can I reproduce it? I see no attached issue.

toberge:
> Pull request looks fine, but what is the crash and how can I reproduce it? I see no attached issue.

Spam fire during loading screens before shooting rounds

ThomasJGabrielsen (approved):

PR #976, opened 2025-05-11
Skips the auction animation after the first round.

### LINKED ISSUE #974: Remove auction start animation in auctions beyond the first
As suggested a long while back

ThomasJGabrielsen:
Implemented by adding an initial empty state to the BiddingCamera and only triggering the animation based on the matchcontroller's RoundCount value

closes #974


toberge (approved):
LGTM!

PR #977, opened 2025-05-12
Prevent items that all players already own to appear in auctions.

### LINKED ISSUE #971: Auction should not sell items that all players already own

ThomasJGabrielsen:
TODO: In the rare case where all items are bought, auction should be skipped, as of now, it simply fails.

closes #971

### DIFF
--- a/Assets/Scripts/Auction/AuctionStage.cs
+++ b/Assets/Scripts/Auction/AuctionStage.cs
@@ -7,6 +7,7 @@ public class AuctionStage : ScriptableObject
     protected float stageDuration = 15;
     [SerializeField]
     protected Item[] items;
+    protected Item[] ownedItems;
 
     /// <summary>
     /// This method should ONLY be called by staticInfo.
@@ -17,6 +18,10 @@ public class AuctionStage : ScriptableObject
         this.items = items;
     }
 
+    public void SetOwnedItems(Item[] Items)
+    {
+        ownedItems = Items;
+    }
     public virtual bool Promote(out BiddingRound round)
     {
         if (items.Length == 0)
--- a/Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
+++ b/Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
@@ -1,3 +1,6 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
 using CollectionExtensions;
 using UnityEngine;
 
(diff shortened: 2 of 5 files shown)


Assets/Scripts/Auction/AuctionStage.cs
```diff
@@ -17,6 +18,10 @@ public void SetItems(Item[] items)
         this.items = items;
     }
 
+    public void SetOwnedItems(Item[] Items)
```
toberge, line 21: Not sure why you're not just setting the items directly? I don't think the auction stages themselves need to know which items are owned?

Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
```diff
@@ -24,23 +27,38 @@ private void OnValidate()
...
+
         for (int i = 0; i < numItems; i++)
         {
-            selection[i] = items[idx[i]];
+            selection[i] = pool[idx[i]];
         }
         return selection;
     }
-
+    private Item[] AvailableSelection()
```
toberge, line 53: If I'm not mistaken, this step could've been happening outside of the stages themselves?

toberge (changes requested):
I agree that we should do this, but I have a few gripes with this implementation. Beyond the comments, I think we should rather auction away a selection of items that have already been picked once we run out of items that haven't, as it'll take a long time to exhaust the item pool and some players might want what some of the others have gotten.

We should also consider weighting items based on whether they have showed up in previous auctions and not been picked, which will require a slightly different approach, but that stuff's for later. (seriously, it'll be fine to just adjust the implementation here and get this in then deal with the other stuff later)

Edit: See #981

Assets/Scripts/Auction/AuctionStage.cs
ThomasJGabrielsen: Auction stages, or rather their inherited classes WeightedRandomisedsedAuctionStage, are currently where we determine what items to add to the auction, thus i felt that they should know which items are owned. I was slightly conflicted if the AuctionStage itself should find and determine what items are owned, but chose instead to let the AuctionDriver.cs do that and pass it along.

Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
ThomasJGabrielsen: I'm a bit uncertain as to what you are referring to here. If it's about determining the available selection outside of each individual auction round (like in the AuctionDriver instead), then passing it to the WeightedRandomisedAuctionStage, then yes that would be an alternative. But the system so far seems to be designed to handle potentially several rounds in the same auction, in which case we should allow each auction stage individually to determine what items can be sold based of changes that happened in previous auction stages.

toberge:
Closing this as it has been implemented otherwise

PR #978, opened 2025-05-12
Disable shadowDecal when player begins to ragdoll.

### LINKED ISSUE #920: Leftover shadow from dead bot?
image

ThomasJGabrielsen:
i.e. when a player dies.

closes #920

### DIFF
--- a/Assets/Prefabs/Input/Player.prefab
+++ b/Assets/Prefabs/Input/Player.prefab
@@ -294,6 +294,7 @@ MonoBehaviour:
   - {fileID: 426369848560938732}
   animatorToDisable: {fileID: 3755527737191155748}
   rigidbodyToPush: {fileID: 4321660441894553452}
+  shadowDecal: {fileID: 5212658928600365944}
 --- !u!82 &7607358150439154286
 AudioSource:
   m_ObjectHideFlags: 0
--- a/Assets/Scripts/RagdollController.cs
+++ b/Assets/Scripts/RagdollController.cs
@@ -1,5 +1,6 @@
 using System.Collections.Generic;
 using UnityEngine;
+using UnityEngine.Rendering.Universal;
 
 public class RagdollController : MonoBehaviour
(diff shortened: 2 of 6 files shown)


Assets/Prefabs/Input/Player.prefab
```diff
@@ -294,6 +294,7 @@ MonoBehaviour:
   - {fileID: 426369848560938732}
   animatorToDisable: {fileID: 3755527737191155748}
   rigidbodyToPush: {fileID: 4321660441894553452}
+  shadowDecal: {fileID: 5212658928600365944}
```
toberge, line 297: hm, I see lots of changes in the derived prefabs, try applying as override to the player prefab in one after removing the change in the rest?

Assets/Scripts/RagdollController.cs
```diff
@@ -1,5 +1,6 @@
 using System.Collections.Generic;
 using UnityEngine;
+using UnityEngine.Rendering.Universal;
```
toberge, line 3: this is required for DecalProjector?

toberge (approved):
Works, but see comments as the prefab changes should only be in one place here (might be slightly wrong)

Assets/Scripts/RagdollController.cs
ThomasJGabrielsen: Yes, DecalProjector inherits from UnityEngine.Rendering.Universal

Assets/Prefabs/Input/Player.prefab
ThomasJGabrielsen: The others are all separate prefabs, we actively use at least 5 different prefabs for the player whether it's AI or not, Tutorial or not, Networked or not, etc.

toberge: Nah they aren't separate actually, they all inherit from Player :))))

PR #983, opened 2025-06-30
Increment version number to 0.5.2

toberge:


Fueredoriku (approved):
It do be like that sometimes

PR #984, opened 2025-07-06
Reorder auctions to barrel, extension, body

toberge:
this makes the most impactful decisions come first, instead of having two bland rounds before getting the chance to pick a different projectile you'll now just have one


toberge:
consider this only after revolver animation softlock has been fixed 😔

Fueredoriku:
Might consider now perchance, but we need to do some extensive testing

PR #985, opened 2025-07-06
Weight auctions based on owned items

toberge:
Requires preparing weights before draws.

wanna add weighting based on which items have been in auctions previously, but that'll take a wee bit more effort

### DIFF
--- a/Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
+++ b/Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
@@ -1,3 +1,5 @@
+using System.Collections.Generic;
+using System.Linq;
 using CollectionExtensions;
 using UnityEngine;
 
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
```diff
@@ -1,42 +1,123 @@
...
 
 #if UNITY_EDITOR
     private void OnValidate()
     {
         numItems = Mathf.Max(1, numItems); // at least one pls
-        if (items.Length != rates.Length)
+        if (items.Length != weightedItems.Count)
         {
-            rates.Resize(items.Length);
+            // TODO eh???? do we need this method???
```
Fueredoriku: Shouldn't need this if the collection is not an array, no. I'm guessing it was an array because of the serializefield (for setting values in the editor)

Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
```diff
@@ -1,42 +1,123 @@
...
     }
 #endif
 
-    private Item[] RandomSelectionWithReplacement()
+    public void PrepareWeights()
     {
-        Item[] selection = new Item[numItems];
-        for (int i = 0; i < numItems; i++)
+        // TODO this shouldn't need to happen for each auction stage
+        //      but relatively inexpensive all things considered...
```
Fueredoriku: Neglible performance-wise to perform once per auction imo

Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
```diff
@@ -1,42 +1,123 @@
...
+                break;
+            }
+            runningWeight += weight;
+        }
+
+        if (index < 0)
+        {
+            Debug.LogWarning($"Found no item at {random} running weight of {totalWeight}");
+            // Default to *something* if we don't find an item
+            (item, weight) = weightedItems.First();
```
Fueredoriku: Could we default to the normal random instead of first maybe?

Fueredoriku (approved):
Looks good, but I haven't been able to test. The code seems correct thought, I just have some comments:

Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
toberge: yeah so I won't refactor that lol

Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
toberge: hm yeah we could, thanks!

toberge: should be a weird edge case that only happens if the code is wrong doe

Fueredoriku: Even so, I think defaulting to something non-deterministic ensures our failsafe is less obvious in case something goes wrong

PR #986, opened 2025-08-17
Weight auctions so unwanted items are less likely

### LINKED ISSUE #981: Weight auctions based on inventories and items not picked
Probable implem:

- Assign score to items
  - Being in x players' inventories lowers its score x * k1
  - Not having been picked in x auctions that it has appeared in lowers its score x * k2
  - k1 > k2 probably
- Pick items using these weights
  - thus, it'll be much less likely but much rarer to end up with less desirable items showing up in auctions

toberge:
Required tracking bought/not bought items.

Closes #981

### DIFF
--- a/Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
+++ b/Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
@@ -1,9 +1,7 @@
 using System.Collections.Generic;
 using System.Linq;
-using CollectionExtensions;
 using UnityEngine;
 
-// TODO: Implement weighted selection with and without replacement
 [CreateAssetMenu(menuName = "Auction/Stage/Weighted Randomised")]
 public class WeightedRandomisedAuctionStage : RandomisedAuctionStage
 {
@@ -12,6 +10,7 @@ public class WeightedRandomisedAuctionStage : RandomisedAuctionStage
     protected int totalWeight;
 
     private Dictionary<Item, int> ownedByXPlayers;
+    private Dictionary<string, int> notBoughtForXRounds;
 
 
 #if UNITY_EDITOR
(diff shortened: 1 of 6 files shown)


Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
```diff
@@ -63,7 +78,17 @@ private int DetermineWeight(Item item)
...
 
-        // TODO check if it has been in previous auction (TODO fix that stuff???)
+        if (notBoughtForXRounds.TryGetValue(item.id, out var roundsNotBought))
+        {
+            score -= roundsNotBought switch
+            {
+                0 => 0,
+                1 => 900,
+                2 => 950,
+                _ => 999,
```
Fueredoriku, line 89: These values might maybe be a bit high? Maybe they should be weighted more similar to ownedByXPlayers?

Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
```diff
@@ -63,7 +78,17 @@ private int DetermineWeight(Item item)
...
+            };
+            Debug.Log($"{item.displayName} not bought last round, scored {score}");
```
Fueredoriku: I think we can remove this

Fueredoriku (approved):
Looks good, but I think the weights should be a bit adjusted

Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
toberge: Yeah, I'd rather add a dedicated debug thingy

Assets/Scripts/Auction/WeightedRandomisedAuctionStage.cs
toberge: Agreed, adjusted the curve 🫡

PR #987, opened 2025-08-18
Clean up guncontroller subscription stuff

toberge:
Took all the not-so-risky fixes from #975 and put them into this PR so we at least have some of these improvements 😔


Assets/Scripts/Gamestate/PlayerManager.cs
```diff
@@ -263,6 +262,8 @@ public void SetPlayerInput(InputManager playerInput)
             var canvas = hudController.GetComponent<Canvas>();
             canvas.worldCamera = inputManager.GetComponentInChildren<Camera>();
             canvas.planeDistance = 0.21f;
+            // TODO pivot inside HUDController (for respawning)
+            // and remember -=
```
Fueredoriku, line 266: Could probably move this comment to hudcontroller if still relevant?

Assets/Scripts/Augment/GunController.cs
```diff
@@ -106,11 +107,31 @@ public void Initialize()
 
     private void OnDestroy()
     {
-        UnsubscribeDelegates();
+        Detach();
     }
 
-    public void UnsubscribeDelegates()
+    // TODO perhaps avoid too many duplacte detach calls?
```
Fueredoriku, line 113: Could probable remove the todo here?

Assets/Scripts/Gamestate/PlayerManager.cs
```diff
@@ -566,17 +543,17 @@ public void SetGunNetwork(Transform offset)
...
         {
             Destroy(gunController.transform.GetChild(i).gameObject);
         }
         Destroy(gunController.gameObject);
-        GunOrigin.GetComponent<GunController>().UnsubscribeDelegates();
         for (int i = GunOrigin.transform.childCount - 1; i >= 0; i--)
         {
             Destroy(GunOrigin.transform.GetChild(i).gameObject);
         }
+        // TODO ok and why don't we destroy the guncontroller for the display gun
```
Fueredoriku, line 556: Might be a good idea to do, is it really not cleared anywhere else? 🤔

Fueredoriku (approved):
looks good, couldnt find any big issues from these changes after testing

Assets/Scripts/Gamestate/PlayerManager.cs
toberge: don't wanna risk touching it now, a bunch of our current code assumes that it is present

Assets/Scripts/Gamestate/PlayerManager.cs
toberge: we don't have a method that takes the player identity in there yet, skipping this for now

Assets/Scripts/Augment/GunController.cs
toberge: No, it's still called from three places and I won't risk weeding them away here

PR #988, opened 2025-08-26
Fix that ducking revolver issue

### LINKED ISSUE #970: Bouncy laser and lightning barrels stop the basic reload on revolver body

toberge:
pls merge the other one first....

or just review this as both of them maybe

Outlining the issue in Norwegian:
- skyt nest siste skudd (her er ammo > 1)
- vent lenge nok til at animasjonen må være ferdig (mens anim kjører er ammo fortsatt > 1)
- prøv å skyte siste skudd (her er ammo == 1 og også <= 1)
--> da får du ikke skutt

This ammo <= 1 check did fix the animation issue we had with you being able to trigger fire animation just as the body started reloading, but removing this not-fix meant I had to fix it otherwise. Adjusted animation speed to actually correspond to the animation length, and gave the laser barrel an actual reduction in fire rate, and now things seem to work smoothly.

Closes #970

### DIFF
--- a/Assets/Scripts/Augment/GunController.cs
+++ b/Assets/Scripts/Augment/GunController.cs
@@ -82,11 +82,6 @@ public class GunController : NetworkBehaviour
 
     private AugmentAnimator barrelAnimator;
 
-    // TODO denne er faktisk ikke i bruk nå siden den ikke blir satt til true lengre.
-    //      -> men hvis du setter den når vi skyter så brekker du spillet
-    //      Fjern denne når/hvis vi har en god erstatning, eller bruk den hvis den funker.
-    private bool isFiring = false;
-
     private int recoilTween;
 
     private void Start()
@@ -173,7 +168,7 @@ public class GunController : NetworkBehaviour
             MatchController.Singleton.onRoundEnd -= CancelZoom;
     }
 
-    private bool ShouldFire() => !isFiring && fireRateController.shouldFire(triggerPressed, triggerHeld);
+    private bool ShouldFire() => fireRateController.shouldFire(triggerPressed, triggerHeld);
 
     [Client]
     private void FixedUpdate()
@@ -291,7 +286,6 @@ public class GunController : NetworkBehaviour
             // hopefully we avoid displaying them in their gruesome nature to the user this way.
             Debug.LogError("Failed to fire gun on owner's client!");
             Debug.LogError(e);
-            isFiring = false;
         }
     }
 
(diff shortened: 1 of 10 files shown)


toberge:
may have some funky results with ddr + lazur

Assets/Scripts/Augment/GunController.cs
```diff
@@ -81,6 +82,8 @@ public void SetPlayerInput(InputManager input)
 
     private AugmentAnimator barrelAnimator;
 
+    // TODO denne er faktisk ikke i bruk nå siden den ikke blir satt til true lengre.
+    //      -> men hvis du setter den når vi skyter så brekker du spillet
```
Fueredoriku: Maybe remove it in that case?

Assets/Scripts/Augment/GunController.cs
```diff
@@ -254,6 +277,7 @@ private void FireGun()
 
         try
         {
+            isFiring = true;
```
Fueredoriku: I suspect something in this block breaks lazur when continously firing with an automatic firemode

Fueredoriku (changes requested):
Very nice to finally fix the revolver!! 😮

Lazur seems really broken now though, using an automatic firemode body like seeker will break the lazur if holding fire contiously. I think this must be fixed before merge

Assets/Scripts/Augment/GunController.cs
toberge: wait shit did I forget to remove that line :skull:

Assets/Scripts/Augment/GunController.cs
toberge: added more to the comment. ideally we'd have a replacement

Assets/Scripts/Augment/GunController.cs
toberge: removed, should work now

Assets/Scripts/Augment/GunController.cs
Fueredoriku: Yeah this seems like something we need to get to the bottom of at some point. My hunch is that there is some animation event firing somewhere that causes the crash

Fueredoriku (approved):
It works! Lets goooo!

Assets/Scripts/Augment/GunController.cs
toberge: Yeah, I'll actually remove this source of confusion I suppose

PR #993, opened 2026-03-28
Offset auction targets by random vector

toberge:
so bidding AI doesn't just go to same bloody spot ugh


toberge:
superseded by #995

